PR #1, opened 2016-10-13
Added Jenkins slave role

robled:
- Local development directories added to .gitignore
- Removed cruft from RHEL6/7 repositories in group_vars and set base_url
  in a generic fashion
- Added Beaker repo where needed
- Found some places where variables/files were named 'jslave'.  Renamed
  to 'jenkins_slave' for consistency.
- Added support for other vagrant providers aside from virtualbox
  (default is still virtualbox)
- The Ansible SSH key can be conveniently used as the Jenkins SSH key
  for development
- certificate_authority role can now be skipped successfully if no CA
  URLs are provided via configuration
- Multiple CA URLs are now supported
- Added "become: true" to various places that needed it
- Updated jswarm version to 2.1
- Created a task to clean yum cache after adding repositories.  Removed
  'update_cache' statements since they are no longer needed.
- Renamed 'jenkins_ssh_private_key' variable for consistency to
  'jenkins_ssh_keyfile'
- Fixed an issue with the 'regenerate private ssh key' handler which was
  named incorrectly
- Added role dependencies for 'jenkins_master' role
- Added a TODO for a possible issue with saving the jenkins-cli.jar file
- Fixed an issue with SSH key permissions
- Fixed an issue with RPM package pinning/blacklisting that could cause
  package installation to be skipped during initial configuration
- Added a TODO for an issue where updating all yum packages can hang
  forever
- Fixed some whitespace issues
- Added a TODO for the python-flake8 package which doesn't appear to be
  available for RHEL6
- Added 'repositories' role in site.yml
- Added ansible.cfg in vagrant directory to resolve SSH connectivity
  issues
- Deleted .vagrant directories that should not be part of the repo
- Added vagrant scripts for Jenkins slave development


group_vars/rhel6
```diff
@@ -1,24 +1,37 @@
...
-    baseurl: "{{ rhel_base }}/latest-RHEL-6/compose/Server/optional/x86_64/os/"
-  - name: rhel6-extras
-    baseurl: "{{ rhel_base }}/latest-EXTRAS-6-RHEL-6/compose/Server/x86_64/os/"
-  - name: rhscl-python27
-    baseurl: https://www.softwarecollections.org/repos/rhscl/python27/epel-6-x86_64/
+  rhscl_python27:
+    name: rhscl-python27
+    baseurl: https://www.softwarecollections.org/repos/rhscl/python27/epel-6-$basearch/
 
 repository_downloads:
```
greg-hellings: Is the conserver repository still necessary? And is the repository_downloads variable used at all anymore?

roles/jenkins_common/tasks/main.yml
```diff
@@ -1,28 +1,30 @@
 - name: install beaker client
+  become: true
```
greg-hellings: If the entire file is become: true, we might decide rather to place become: true at the level of site.yml for this. But we'd have to be OK with that propagating down to dependency roles included from meta/main.yml

roles/jenkins_master/meta/main.yml
```diff
@@ -0,0 +1,4 @@
+dependencies:
+  - role: certificate_authority
+  - role: repositories
+  # TODO: consider adding jenkins_common role as a dependency
```
greg-hellings: Is this still a todo?

roles/jenkins_slave/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
+- name: install rpm packages
+  yum:
+    name: "{{ jslave_rpm_deps | join(',') }}"
+    state: present
+
+- name: install pip packages
```
greg-hellings, line 6: It would be nice if we could eventually move this to not be using pip to install globally. That's generally a big no-no, when Python is handled by an external package manager like ours is. Then you have pip and yum competing over certain files.

Obviously this is how it's already done, but it's something for us to consider going forward.

greg-hellings (commented):
Once this is merged, let's create issues to track the TODO items listed throughout the code.

roles/jenkins_master/meta/main.yml
robled: It is, we would need to modify the jenkins_master role to use the jenkins_common role.  I was thinking we could defer that work into a separate issue since it may require a decent amount of modification and testing to make happen.

roles/jenkins_slave/tasks/main.yml
robled: Agreed, this is less than ideal, but it's the best we can do for now and is part of the port from the previous version of this software.  If you have an idea to make this better we can discuss it offline and turn it into an actionable ticket.

group_vars/rhel6
robled: As far as I know consvr is still in use according to the previous project (targets/repos.py).  repository_downloads is not used elsewhere, I'll remove that.

roles/jenkins_common/tasks/main.yml
robled: I think that become:true as default best suits our use case, where it is needed the large majority of the time.  If you agree with this I can make the necessary modfications to put this in site.yml and re-test.  Let me know.

group_vars/rhel6
greg-hellings: As we discussed directly, if conserver is not a necessity for the Jenkins function let's remove this bogus URL from this file so it doesn't produce potential breakage in the future. If any users want to use conserver, then they can add the repository through their group_vars as intended.

roles/jenkins_common/tasks/main.yml
greg-hellings: I like to remove as much repetition from playbooks as possible. I'd prefer to see this at the higher level. If the level of efort is significant, let's open an issue on it rather than do it here.

roles/jenkins_master/meta/main.yml
greg-hellings: Opening an issue is my thought as well, I just wanted to be sure that we actually need this.

roles/jenkins_slave/tasks/main.yml
greg-hellings: Definitely separate issue type of work. It would be best, I would think, to put this into a virtualenv instead of into the system root, along with the two packages below this one.

PR #2, opened 2016-11-07
Packaging

greg-hellings:
Added setup.py to allow for packaging of files.


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+  </component>
+</project>
```
robled: Is this directory/file intended to be part of this commit?  I noticed that .idea is part of the .gitignore list.

setup.py
```diff
@@ -0,0 +1,37 @@
...
+from setuptools import setup, find_packages
+from os import path
+
+here = path.abspath(path.dirname(__file__))
+with open(path.join(here, 'README.md')) as f:
+    description = f.read()
+
+setup(
+    name='cinch',
+    version='0.1.6',
```
robled, line 12: Does this follow [semantic versioning](http://semver.org/)?  It looks like it, but I just want to be sure.

.idea/vcs.xml
greg-hellings: Ah, one of them sneaked in. I'll excise this.

setup.py
greg-hellings: I feel like it is. We're declaring it pre-alpha, it's pre-1.0 versioning. We haven't even made a release yet. And I've just done a lot of iterating on what I would call a 0.1 release here. The .6 came up just as I was testing updating and upgrading it in-place. Our first "release" should probably be called 0.2.0 just to keep things sane.

robled (approved):

PR #6, opened 2016-12-14
Initial support for linchpin integration

robled:
* Made 'os' import statements consistent between entry_point.py and
wrappers.py

* Fixed some line length issues in cinch() in entry_point.py

* Added 'cinchpin' command and associated CLI arguments.  This command can
be called the same way as 'linchpin', e.g. "cinchpin rise" or optionally
you can provide a path to the linch-pin working directory, e.g.
"cinchpin rise -w /path/to/workdir".

* Added "pushd" work-around using a contextmanager to allow linch-pin to
be in the correct working directory to operate

* Fixed some line length issues in call_ansible()

* Added StrictHostKeyChecking=no as a default setting to aid in
automation with call_ansible().  This removes the need for ansible.cfg
entirely.

* Created a generic command_handler() function so that call_ansible()
and call_linchpin() can share some code.

* Added call_linchpin() to find necessary data within in the linch-pin
working directory so that call_ansible can be passed the necessary
inventory file when it is called after provisioning.  This function
supports a subset of linchpin commands (currently only the 'rise'
subcommand is available, but more will be added in the future).

* Added an Ansible role named "check_ssh" that is a work-around to
handle waiting on cloud-init to create a working SSH configuration on
the server side.  Without this, Ansible will fail pretty spectactularly
and surprisingly doesn't have a built-in way to simply retry SSH
connections until authentication is successful.


cinch/bin/wrappers.py
```diff
@@ -13,30 +15,115 @@
...
+    :param work_dir: The linch-pin working directory that contains a PinFile
+    and associated configuration files
+    :param args: An array of other command-line arguments to linchpin to pass
+    :return: The exit code returned from linchpin, or 255 if errors come from
+    elsewhere
+    """
+    # Construct the arguments to pass to linchpin by munging the arguments
+    # provided to this method
+    linchpin_args = list()
+    linchpin_args.extend(args)
```
greg-hellings: Why did you break this up into two calls, instead of just creating a copy of args? There's a few ways to do it, but the most simple would be
linchpin_args = args[:]
or
linchpin_args = list(args)

cinch/bin/wrappers.py
```diff
@@ -13,30 +15,115 @@
...
+    # cinch will only support a subset of linchpin subcommands
+    supported_cmds = ['rise']
+    for cmd in linchpin_args[0]:
```
greg-hellings: Is the intention to eventually support executing multiple linchpin commands at once, e.g. to be able to do
cinchpin rise fall
and do the full run of provision/configure/teardown?

I'm curious why you're looping on a single element of linchpin_args here.

cinch/bin/wrappers.py
```diff
@@ -13,30 +15,115 @@
...
+        chdir(previous_dir)
+
+    with pushd(work_dir):
+        linchpin = local['linchpin']
+        exit_code = command_handler(linchpin, linchpin_args)
+    # If linchpin is asked to provision resources, we will then run our
+    # cinch playbooks
+    # TODO: Add support for 'drop' and other supported linchpin command
+    # subsets
+    if 'rise' in linchpin_args[0] and exit_code == 0:
```
greg-hellings: Do you want to use "in" instead of "==" for the operator, here? "in" will return true for other strings there such as "arisen" or such. Not that those are valid linchpin commands, but in the future they might be.

cinch/roles/check_ssh/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+ansible_port: 22
+ssh_retries: 100
+ssh_retry_delay: 6
+host: "{{ hostvars[inventory_hostname]['ansible_host'] | default(inventory_hostname) }}"
```
greg-hellings, line 3: hostvars[inventory_hostname] is probably redundant here. By default, values like "ansible_host" are pulled from the current host's variables. I don't object to being explicit, here, as this is intentionally kludgy code to work around Ansible's internal aversion to recovery from SSH host unreachable. I just wanted you to be aware of that.

cinch/roles/check_ssh/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+ansible_port: 22
```
greg-hellings: With our change from raw SSH to the ansible ping kludge, is this necessary? I don't see it being used, but I could be missing it, somewhere.

cinch/bin/wrappers.py
robled: This code has been removed since we are only supporting a single linch-pin argument.

cinch/bin/wrappers.py
robled: As previously discussed I can see no reason to support multiple linch-pin arguments at this time, so this has  been removed.

cinch/bin/wrappers.py
robled: This code has been removed since we are only supporting a single linch-pin argument.

cinch/roles/check_ssh/defaults/main.yml
robled: Removed as it is unused.

cinch/roles/check_ssh/defaults/main.yml
robled: I've decided to keep this for the purposes of being explicit.

greg-hellings (approved):
Looks ready for primetime to me.

PR #10, opened 2017-01-12
"init" support added

robled:
* Imported the entire 'os' module since it is used often and updated
references to it

* Fixed some outdated docstrings

* Moved inventory file validation checking and loading to a separate
function for readability

* Added a configuration skeleton for ease of use

* Elected to overwrite the example PinFile provided by linch-pin since
the example contains many sections that we do not use

* Improved error handling for topology files

### DIFF
--- a/cinch/bin/wrappers.py
+++ b/cinch/bin/wrappers.py
@@ -6,11 +6,18 @@ from plumbum.commands.processes import ProcessExecutionError
 from traceback import print_exc
 
 import os
-import yaml
 import sys
+import yaml
 
 
 BASE = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
+# Link to our docs with configuration examples
+DOCS = 'http://example.com'
+# Skeleton text to insert in YAML config files
+SKEL_TEXT = '''---
+# Add your cinch {0} configuration here
+# Examples: {1}
+'''
 
 
 def call_ansible(inventory, playbook, *args):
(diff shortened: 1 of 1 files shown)


cinch/bin/wrappers.py
```diff
@@ -3,14 +3,14 @@
 from contextlib import contextmanager
 from plumbum import local
 from plumbum.commands.processes import ProcessExecutionError
-from os import path, chdir, getcwd
```
greg-hellings, line 6: Why'd you make this change?

cinch/bin/wrappers.py
```diff
@@ -44,19 +44,100 @@ def call_linchpin(work_dir, arg):
...
     :return: The exit code returned from linchpin, or 255 if errors come from
     elsewhere
     """
     # cinch will only support a subset of linchpin subcommands
-    supported_cmds = ['rise', 'drop']
+    supported_cmds = ['rise', 'drop', 'init']
     if arg not in supported_cmds:
         raise Exception('linchpin command "{0}" not '
                         'supported by cinch'.format(arg))
+
```
greg-hellings, line 63: This method feels like it's complicated enough we should start thinking about factoring it out into some composing functions that just get selected based on if logic.

cinch/bin/wrappers.py
```diff
@@ -44,19 +44,100 @@ def call_linchpin(work_dir, arg):
...
+                'layout': config_file
+            }
+        }
+        # These are the first-level config paths that linch-pin creates for us
+        # in the working directory
+        local_paths = ['layouts', 'topologies']
+        # Link to our docs with configuration examples
+        docs = 'http://example.com'
+        # Skeleton text to insert in YAML config files
+        skel_text = '''---
```
greg-hellings: Do we have a standard for putting multi-line strings outside of the flow (e.g. as file globals)? If not, I propose that we do so, as these lines interrupt the normal flow of Python code. We could even factor them out into a separate file if we end up with multiple such strings.

greg-hellings (changes requested):

cinch/bin/wrappers.py
robled: Mainly for convenience.  As discussed we will leave it be and just import all of the 'os' module.

cinch/bin/wrappers.py
robled: The 'init' code has been moved into a separate function now.

cinch/bin/wrappers.py
robled: These have been moved to the top of the file.

PR #12, opened 2017-01-19
initial set of sphinx docs

robled:
* Converted existing markdown files to restructured text

* Created a directory layout and configuration for sphinx


docs/source/conf.py
```diff
@@ -0,0 +1,153 @@
...
+project = u'cinch'
+copyright = u'2017, RedHatQE'
+author = u'RedHatQE'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = u''
```
greg-hellings, line 57: I feel like there should be a way for us to automate pulling the version out of setup.py and into here at some point in the future, when we're automating the release process. Totally not necessary right now, but let's keep this line and line 59 in mind when we get to that point.

greg-hellings (approved):

Issue #21, opened 2017-02-07
NTP


greg-hellings:
Ensure that NTP is installed and operating on Jenkins masters

PR #41, opened 2017-02-10
Add NTP role

greg-hellings:
Added NTP role to jenkins_master and updated site.yml

Addresses #21

### DIFF
--- a/cinch/site.yml
+++ b/cinch/site.yml
@@ -29,6 +29,7 @@
   hosts: jenkins_master
   roles:
     - jenkins_master
+    - ntp
 
 - name: configure jenkins slaves
   become: true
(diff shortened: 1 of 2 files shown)


cinch/site.yml
```diff
@@ -29,6 +29,7 @@
   hosts: jenkins_master
   roles:
     - jenkins_master
+    - ntp
```
robled, line 32: Would there be any downside to having Jenkins slaves also include the NTP role?  I know that Jenkins can get upset if slaves have clocks that are out of sync with the master.

greg-hellings: I don't know if our goal is to keep slaves as tightly controlled as possible with minimal install other than Swarm or not. We previously have had ntp running on masters and not on slaves, but I put it in a separate role so that it can easily be added to slaves if we want. I could go either way on that.

robled: Fair point, if someone needs NTP on a Jenkins slave they can always include it.  I like the "minimal" concept.

Issue #30, opened 2017-02-07
Improve Plugin Install


greg-hellings:
Improve installing Jenkins plugins to directly call the API instead of passing calls through the Jenkins CLI jar file.

greg-hellings:
Making this improvement should also allow specific versions of plugins to be installed per the install files, rather than forcing the latest to be installed.

greg-hellings:
Since Ansible 2.2, there exists a core module to manage Jenkins plugins. We should switch to that and ditch the custom installer.

PR #42, opened 2017-02-14
Jenkins plugin install improvements

greg-hellings:
Removed dependence on jenkins-cli.jar
Added patched jenkins_plugin.py to library

Addresses #30


cinch/library/jenkins_plugin.py
```diff
@@ -0,0 +1,826 @@
+#!/usr/bin/python
```
robled, line 1: I know that this file was brought in from another repo, but should we change the shebang to **#!/usr/bin/env python** to match the rest of our code?  Also, would a git submodule be appropriate here?  Just curious.

greg-hellings: I wanted to keep the file as close to upstream as possible, so I didn't change anything other than the necessary lines for bug fixing. Changing the shebang wouldn't be a problem, since Ansible modules get directly executed by calling "python some_module.py"

Do you mean a Python submodule or a git submodule?

robled: I meant a git submodule.  I take it you did perform some modification of the original code?  If that's the case then we can leave it as-is.  I thought it was brought in wholesale from another repo.

greg-hellings: There were two bugs that caused the module to error out installing some of the plugins. I've modified them in this version, and I'll be putting in a PR against upstream if they haven't been resolved there, already.

robled (approved):

cinch/library/jenkins_plugin.py
greg-hellings: Noting here: Ansible 2.2.1 includes a version of jenkins_plugin.py that addresses this bug, along with a few other minor issues. I am going to create a separate issue to drop this file once we upgrade to a version of linchpin that depends on Ansible 2.2.1, and can be guaranteed of the fixed version of Ansible being in our path.

PR #47, opened 2017-02-22
Systemd upstart

greg-hellings:
Convert jswarm service to use Upstart (CentOS 6/RHEL 6) or systemd (CentOS 7/RHEL 7).


cinch/roles/jenkins_slave/tasks/main.yml
```diff
@@ -36,42 +36,43 @@
...
+  notify: restart swarm
 
-- name: install swarm to rc.local
-  # TODO: replace with upstart file and systemd service
-  lineinfile:
-    dest: /etc/rc.d/rc.local
+- name: upload swarm systemd file
+  template:
+    src: swarm.service
+    dest: /usr/lib/systemd/system/swarm.service
```
robled: Should this go in **/etc/systemd/system** since the service is not provided by a package?

greg-hellings: I have no knowledge of that folder. I can test it out, if you'd like.

robled: Yeah, let's go that route.  Most of the docs I've read about systemd indicate that's where user-provided service files should live.  You might need to reload systemd to get it to pick up changes if you're not doing that already.

greg-hellings:
I moved the systemd files to /etc/systemd/system, but also discovered in the process I had missed one gating condition on the upstart file.

cinch/roles/jenkins_slave/templates/swarm.upstart.conf
```diff
@@ -0,0 +1,21 @@
+description "Jenkins slave swarm client"
```
robled, line 1: Something I missed here the first time around is that this template file does not end with a .j2 extension.  Is this a standard convention we should be following?

greg-hellings: Ansible convention seems to go with the .j2 extension but I avoid it. With the .j2 extension basically no text editor around will pick up proper syntax highlighting for the file. I've found that many of them will still get the highlighting correct if you just leave off the .j2 extension. (It's not foolproof with vim, since some files are picked up by their location on the filesystem, rather than extension, but it has a much greater chance of success than with the .j2 extension - therefore, I always eschew the extension)

greg-hellings: I could go either way, I just got in the habit of leaving off the .j2 because it doesn't seem helpful that I can tell. It's already possible to discern that the files are templates based on the directory structure.

robled: Good point about the syntax highlighting.  I prefer to have that working rather than following Ansible's conventions.

PR #53, opened 2017-02-28
use service module to stop swarm client on teardown #52

robled:
#52


greg-hellings (approved):
D'oh. I kept reminding myself to make this change and then completely forgot it. Good catch.

Issue #48, opened 2017-02-23
Compatibility issues with jenkins 2?


jjanco:
My playbook.yml:
- hosts: default
  vars:
    some certificates and repositories hidden
    gcc_compat_package: compat-gcc-44
    jenkins_rpm: jenkins
    jenkins_ssl_cert: cinch/vagrant/master_ssl/jenkins.crt
    jenkins_ssl_key: cinch/vagrant/master_ssl/jenkins.key
    jenkins_security_enabled: true
  roles:
    - certificate_authority
    - repositories
    - jenkins_common
    - jenkins_master

Test machine: centOs7
jenkins_rpm ~ jenkins-2.32

First error:
TASK [jenkins_master : download jenkins CLI] ***********************************
fatal: [default]: FAILED! => {"changed": false, "failed": true, "msg": "Destination /var/lib/jenkins not writable"}

Workaround with permissions 777

Errors about configuring users:

TASK [jenkins_master : configure CLI users appropriately] **********************
fatal: [default]: FAILED! => {"changed": false, "failed": true, "msg": "Roles not found - have you configured an admin using the Role-based Authorization Strategy?"}

Workaround with  jenkins_security_enabled: false

Errors about plugins:
TASK [jenkins_master : install plugins] ****************************************
FAILED - RETRYING: TASK: jenkins_master : install plugins (3 retries left).
FAILED - RETRYING: TASK: jenkins_master : install plugins (2 retries left).
FAILED - RETRYING: TASK: jenkins_master : install plugins (1 retries left).
failed: [default] (item=build-name-setter) => {"attempts": 3, "details": "HTTP Error 403: Forbidden", "failed": true, "item": "build-name-setter", "msg": "Cannot get CSRF"}
FAILED - RETRYING: TASK: jenkins_master : install plugins (3 retries left).
FAILED - RETRYING: TASK: jenkins_master : install plugins (2 retries left).
FAILED - RETRYING: TASK: jenkins_master : install plugins (1 retries left).
failed: [default] (item=envinject) => {"attempts": 3, "details": "HTTP Error 403: Forbidden", "failed": true, "item": "envinject", "msg": "Cannot get CSRF"}
FAILED - RETRYING: TASK: jenkins_master : install plugins (3 retries left).
FAILED - RETRYING: TASK: jenkins_master : install plugins (2 retries left).
FAILED - RETRYING: TASK: jenkins_master : install plugins (1 retries left).
[...]

PR #54, opened 2017-03-01
Skip Jenkins 2 initial setup

kubco2:
After installation of Jenkins 2, there is initial setup page and Jenkins is installed with initialPassword, then it is not possible to configure users and plugins  #48.
With added jenkins java option this setup is skipped and Jenkins is installed unsecured, like Jenkins 1.6x.


robled:
Is this change also compatible with Jenkins 1.6.x?

kubco2:
I tried run of playbook - installation of Jenkins 1.6 and test if webif works. Both worked, I didnt dig into it more.

greg-hellings:
We can use version_compare to be sure we only apply the entry to
installations of Jenkins >= 2 to avoid possible problems.

https://docs.ansible.com/ansible/playbooks_tests.html#version-comparison

scoheb:
No issues with this in Jenkins 1.651.X. Currently in use by RH in provisioner 1.0 and puppet.

PR #60, opened 2017-03-20
Jswarm container

greg-hellings:
Please bear with me while I work on the TravisCI stuff. Might not build properly the first go.


cinch/roles/dockerize/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+    dest: /usr/local/bin
+  when: "{{ dockerize_download | changed }}"
+
+- name: make sure dockerize has proper attributes
+  file:
+    dest: /usr/local/bin/dockerize
+    mode: 0755
+    owner: root
+    group: root
+    state: file
```
robled, line 20: Does this task run solely in Travis?

cinch/roles/jenkins_slave/tasks/main.yml
```diff
@@ -36,6 +37,14 @@
     dest: "{{ jenkins_user_home }}/{{ jswarm_filename }}"
     url: "{{ jswarm_download }}"
 
+- name: set variable for docker container
+  set_fact:
+    jenkins_master_url: "{{ '{% raw %}{{ .Env.JENKINS_MASTER_URL }}{% endraw %}' }}"
```
robled: I know that this templating stuff was tricky to get working.  This might deserve a comment just to explain why it was required and what the expected end result should be.

cinch/roles/jenkins_slave/templates/sysconfig_jenkins_swarm
```diff
@@ -6,4 +10,4 @@ SWARM_SLAVE_LABEL="{{ jslave_label }}"
 SWARM_ROOT="{{ jswarm_home }}"
 SWARM_USERNAME="{{ (jenkins_slave_username == '') | ternary('', '-username') }} {{ jenkins_slave_username }}"
 SWARM_PASSWORD="{{ (jenkins_slave_password == '') | ternary('', '-password') }} {{ jenkins_slave_password }}"
-SWARM_EXTRA_ARGS="-mode exclusive -disableSslVerification -deleteExistingClients"
+SWARM_EXTRA_ARGS="-mode exclusive -disableSslVerification -deleteExistingClients {{ jswarm_extra_args }}"
```
robled, line 13: What limitation prevents us from sharing the generated swarm CLI between the three places that need it? Just curious.

scripts/cent6_jswarm.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -e
+
+CINCH="$(readlink -f "$(dirname "$0")/../")"
+CINCH_VERSION=$(grep "${CINCH}/setup.py" -e 'version=' | sed -e "s/.*version='\(.*\)'.*/\1/")
```
robled: Add a comment for this regex.

scripts/cent6_jswarm.sh
```diff
@@ -0,0 +1,18 @@
...
+
+echo "*****************************************************"
+echo "Building cinch container for version ${CINCH_VERSION}"
+echo "*****************************************************"
+
+echo "Starting container"
+ansible -i /dev/null localhost -m docker_container -a "image=greghellings/centos_base:6 name=jswarm detach=true tty=true command=/bin/bash"
```
robled: Does Docker hub support team management of accounts?  This would be preferable to using a personal account to store the generated image.

scripts/cent7_jswarm.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -e
+
+CINCH="$(readlink -f "$(dirname "$0")/../")"
+CINCH_VERSION=$(grep "${CINCH}/setup.py" -e 'version=' | sed -e "s/.*version='\(.*\)'.*/\1/")
```
robled: Add a comment for this regex.

scripts/cent7_jswarm.sh
```diff
@@ -0,0 +1,18 @@
...
+
+echo "Starting container"
+ansible -i /dev/null localhost -m docker_container -a "image=greghellings/centos_base:latest name=jswarm detach=true tty=true command=/bin/bash"
+echo "Building container with Ansible"
+ansible-playbook -i "${CINCH}/inventory/cent7_jswarm_docker" "${CINCH}/cinch/site.yml" \
+	-e jenkins_user_password=some_dummy_value
+echo "Committing container at tag ${CINCH_VERSION}"
+docker commit \
+	--change 'ENTRYPOINT ["/usr/local/bin/dockerize", "-template", "/etc/sysconfig/jenkins_swarm:/etc/sysconfig/jenkins_swarm.templated", "/opt/jswarm.sh"]' \
+	jswarm "cinch:cent7-${CINCH_VERSION}"
```
robled: This file looks pretty similar to the cent6 file.  Can we have the script take a parameter for the necessary differences?

robled (changes requested):

cinch/roles/dockerize/tasks/main.yml
greg-hellings: No, it runs solely inside of a Docker container. The "dockerize" utility allows inserting environment variables into a file and templating it out. For machine-based installs, we template values directly into the config files with Ansible, but with Docker image builds we need to allow the user to supply those values at run time instead of at configure time.

cinch/roles/jenkins_slave/tasks/main.yml
greg-hellings: I've added a War and Peace style comment to explain this very odd syntax.

cinch/roles/jenkins_slave/templates/sysconfig_jenkins_swarm
greg-hellings: The main reason I originally moved away from the "construct the full command in Ansible" method when we moved to systemd/upstart was because doing so left no trace on the filesystem of an easily editable configuration file. Someone who was tweaking an on-disk running slave would need to edit the systemd (or upstart or shell) file directly instead of being able to edit a configuration file.

Maintainability of this software argues for moving back to constructing the string in Ansible, but there are other technical difficulties there. Such as
* How do we properly account for optional -password and -username args in a Docker container, where we don't know if they're supplied until runtime?
* How do we make it easy on a user to edit a configuration file to test changes to their system in real time?

scripts/cent6_jswarm.sh
greg-hellings: In this case, I'm not storing the resulting image in my personal account, rather I'm using one of my personal images as a base. That image contains CentOS with yum update and yum install sudo already executed. Doing this allows me to avoid installing sudo from Ansible, where it's just spurious noise that is already present in all machine based CentOS installs.

Of course, I could also move the installation of such software into this script using ansible ad hoc commands. I had done things this way because my home Internet connection is incredibly slow, so I wanted to have as much of the software pre-installed as possible. On the hosts (e.g. Travis, Jenkins, etc) that we're going to be using, such bandwidth limitations are almost non-existent, so I'm not at all opposed to moving this to a base CentOS image pull.

scripts/cent7_jswarm.sh
greg-hellings: Comment added

scripts/cent7_jswarm.sh
greg-hellings: I wasn't sure if they would grow more and diverge in the future. I've opted for combining them into a single centos_jswarm.sh file that takes an argument to specify what it wants to build.

cinch/roles/dockerize/tasks/main.yml
robled: Understood, thanks!

cinch/roles/jenkins_slave/templates/sysconfig_jenkins_swarm
robled: Fair points, and thanks for the explanation.

cinch/roles/jenkins_slave/tasks/main.yml
```diff
@@ -37,6 +37,13 @@
     dest: "{{ jenkins_user_home }}/{{ jswarm_filename }}"
     url: "{{ jswarm_download }}"
 
+# Inside of a Docker container, we want to be able to pull values from the container's running
+# environment. In order to do that, we need to drop values into the configurable fields that
+# match the syntax dockerize uses. Unfortunately, that syntax is very similar to Jinja2 syntax,
+# and Ansible does not stop evaluation after only one iteration. Putting Jina2 raw blocks into
```
robled: typo "Jina2"

scripts/centos_jswarm.sh
```diff
@@ -0,0 +1,28 @@
...
+
+CINCH="$(readlink -f "$(dirname "$0")/../")"
+CENTOS_VERSION="${1}"
+INVENTORY="${CINCH}/inventory/cent${CENTOS_VERSION}_jswarm_docker"
+if [ ! -e "${INVENTORY}" ]; then
+	echo "You must specify a CentOS version to continue"
+	exit 1
+fi
+# Extracts the version line from the setup.py script, and trims off the rest of the line to leave
+# only the expected version
```
robled, line 11: Please add an example of what the expected version string should look like in this comment.

scripts/centos_jswarm.sh
```diff
@@ -0,0 +1,28 @@
...
+CINCH_VERSION=$(grep "${CINCH}/setup.py" -e 'version=' | sed -e "s/.*version='\(.*\)'.*/\1/")
+
+echo "*****************************************************"
+echo "Building cinch container for version ${CINCH_VERSION}"
+echo "*****************************************************"
+
+echo "Starting container"
+ansible -i /dev/null localhost -m docker_container -a "image=centos:${CENTOS_VERSION} name=jswarm detach=true tty=true command=/bin/bash"
```
robled: This line is pretty long, can we wrap it?

scripts/centos_jswarm.sh
```diff
@@ -0,0 +1,28 @@
...
+ansible -i "${INVENTORY}" all -m yum -a "name=sudo state=present"
+ansible -i "${INVENTORY}" all -m yum -a "name=* state=latest"
+echo "Building container with Ansible"
+ansible-playbook -i "${INVENTORY}" "${CINCH}/cinch/site.yml" \
+	-e jenkins_user_password=some_dummy_value
+echo "Committing container at tag ${CINCH_VERSION}"
+docker commit \
+	--change 'ENTRYPOINT ["/usr/local/bin/dockerize", "-template", "/etc/sysconfig/jenkins_swarm:/etc/sysconfig/jenkins_swarm.templated", "/opt/jswarm.sh"]' \
```
robled: This line is pretty long, can we wrap it?

cinch/roles/jenkins_slave/templates/sysconfig_jenkins_swarm
greg-hellings: I think it's worth brainstorming and discussing how we might reduce the maintenance burden on this particular point.

scripts/centos_jswarm.sh
greg-hellings: It literally can be anything. It's .* in the regex. I don't think, technically, there's any limitations on what the version string can be (we're following best practices and going with semver for now, but there's no need or requirement that we do, and this code supports literally anything).

scripts/centos_jswarm.sh
greg-hellings: If someone knows the Bash syntax for that better than I do, please feel free to add it!

robled (approved):

PR #65, opened 2017-04-01
Jswarm container tweaks

greg-hellings:
As I've been messing more and more with OpenShift I've been running into a few problems with the built image. Although the image works fine in raw Docker, it doesn't always seem to live up to the security standards of OpenShift.

Also, noticed a few file permission errors and a hard-coded instance of the Jenkins username in the sudoers file.

### DIFF
--- a/cinch/roles/jenkins_slave_container_wrapup/templates/jswarm.sh
+++ b/cinch/roles/jenkins_slave_container_wrapup/templates/jswarm.sh
@@ -1,5 +1,5 @@
 #!/bin/bash -xe
-. /etc/sysconfig/jenkins_swarm.templated
+. /var/lib/jenkins/jenkins_swarm.templated
 exec java ${SWARM_JAVA_ARGS} \
 	-jar "{{ jenkins_user_home }}/{{ jswarm_filename }}" \
 	-master "${SWARM_MASTER}" \
--- a/cinch/roles/jenkins_slave_container_wrapup/templates/jswarm.sh
+++ b/cinch/roles/jenkins_slave_container_wrapup/templates/jswarm.sh
@@ -1,7 +1,7 @@
 #!/bin/bash -xe
-. /var/lib/jenkins/jenkins_swarm.templated
+. /etc/sysconfig/jenkins_swarm.templated
 exec java ${SWARM_JAVA_ARGS} \
-	-jar "{{ jenkins_user_home }}/{{ jswarm_filename }}" \
+	-jar "{{ jswarm_local_directory }}/{{ jswarm_filename }}" \
 	-master "${SWARM_MASTER}" \
 	-name "${SWARM_SLAVE_NAME}" \
 	-executors "${SWARM_EXECUTORS}" \
--- a/scripts/centos_jswarm.sh
+++ b/scripts/centos_jswarm.sh
@@ -31,5 +31,5 @@ ansible-playbook -i "${INVENTORY}" \
 echo "Committing container at tag ${CINCH_VERSION}"
 docker commit \
 	--change 'USER jenkins' \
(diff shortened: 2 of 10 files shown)


scripts/centos_jswarm.sh
```diff
@@ -30,5 +30,6 @@ ansible-playbook -i "${INVENTORY}" \
 
 echo "Committing container at tag ${CINCH_VERSION}"
 docker commit \
-	--change 'ENTRYPOINT ["/usr/local/bin/dockerize", "-template", "/etc/sysconfig/jenkins_swarm:/etc/sysconfig/jenkins_swarm.templated", "/opt/jswarm.sh"]' \
+	--change 'USER jenkins' \
+	--change 'ENTRYPOINT ["/usr/local/bin/dockerize", "-template", "/etc/sysconfig/jenkins_swarm:/var/lib/jenkins/jenkins_swarm.templated", "/opt/jswarm.sh"]' \
```
robled: There might be some convenience/debugging value in putting the startup script in **/usr/local/bin**, since that will be part of $PATH.

cinch/roles/jenkins_slave_container_wrapup/templates/jswarm.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash -xe
-. /etc/sysconfig/jenkins_swarm.templated
+. /var/lib/jenkins/jenkins_swarm.templated
```
robled: This might be better placed in **/opt** or **/usr/local/bin**.  I think that **/var/lib** should be reserved for the package manager's use.

robled (changes requested):

cinch/roles/jenkins_slave_container_wrapup/templates/jswarm.sh
greg-hellings: /var/lib/jenkins is our $HOME for the jenkins user. The dockerize command has to execute as the jenkins user for use in OpenShift, so we need to have the location of the jenkins_swarm.templated file be writable by that user.

I defaulted to putting it in that user's home directory, but it's definitely not the only viable choice. We could pre-create an empty file at /etc/sysconfig/jenkins_swarm.templated that is owned by jenkins in the container cleanup role and leave the file there, or we could change permissions on some other folder for a container as well. Doing so would save us problems if we ever change the default value of {{ jenkins_user_home }}.

scripts/centos_jswarm.sh
greg-hellings: Are you referring to the /opt/jswarm.sh or is this paralleling the comment about the config file below?

robled: In this case I'm referring to **/opt/jswarm.sh**.

cinch/roles/jenkins_slave_container_wrapup/templates/jswarm.sh
robled: You raise a good point about changing {{ jenkins_user_home }}.  We should strive to survive that scenario at least.  As for the path where we put this file, let's just use our best judgement and follow the [FHS](http://www.pathname.com/fhs/pub/fhs-2.3.html).

greg-hellings: If I read that correctly, then we definitely should put this config file back into /etc and munge with permissions there, as necessary, to drive the execution of that.

But we also should move the default location of the JSwarm .jar file to /opt/jswarm as well, and the shell script could go into /opt/jswarm/bin? Or do you still think /usr/local/bin is better for that?

robled: Yeah, let's use **/etc** for the config file and make necessary permissions changes.  **/opt/jswarm** is good for the jar file, and I would put the startup script into **/usr/local/bin**.

robled (approved):

PR #66, opened 2017-04-07
roles/jenkins_slave: support inject of env vars

thrix:
This patch provides a convient way how to add additional
environment variables to a running jswarm client.

Signed-off-by: Miroslav Vadkerti <mvadkert@redhat.com>

### DIFF
--- a/cinch/roles/jenkins_slave/defaults/main.yml
+++ b/cinch/roles/jenkins_slave/defaults/main.yml
@@ -7,6 +7,8 @@ jenkins_url: http://example.com:8080/jenkins/
 jslave_name: "{{ inventory_hostname }}"
 # The label for this slave
 jslave_label: ops-jslave
+# Additional list of variables injected into jswarm's environment
+jswarm_environment: []
 # Number of executor processes on the slave - modify based on system capacity and load
 jswarm_execs: 10
 # Extra command line args passed to jswarm
--- a/cinch/roles/jenkins_slave/tasks/main.yml
+++ b/cinch/roles/jenkins_slave/tasks/main.yml
@@ -69,6 +69,14 @@
     mode: 0644
   notify: restart swarm
 
+- name: inject additional variables to swarm's environment
+  lineinfile:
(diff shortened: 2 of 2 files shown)


cinch/roles/jenkins_slave/defaults/main.yml
```diff
@@ -7,6 +7,8 @@ jenkins_url: http://example.com:8080/jenkins/
 jslave_name: "{{ inventory_hostname }}"
 # The label for this slave
 jslave_label: ops-jslave
+# Additional list of variables injected into jswarm's environment
+jswarm_environment: []
```
greg-hellings, line 11: Can you give a little documentation here about the format these variables are supposed to be expressed in?

cinch/roles/jenkins_slave/tasks/main.yml
```diff
@@ -69,6 +69,14 @@
     mode: 0644
   notify: restart swarm
 
+- name: inject additional variables to swarm's environment
+  lineinfile:
+    dest: /etc/sysconfig/jenkins_swarm
+    line: "{{ item }}"
+    state: present
+  with_items: "{{ jswarm_environment }}"
+  notify: restart swarm
```
greg-hellings, line 78: Rather than upload a file, then modify it, can you move this into the jenkins_swarm template file and have it iterate over the list of variables directly in the template?

greg-hellings (changes requested):

cinch/roles/jenkins_slave/defaults/main.yml
thrix: Sure, will do in an update

cinch/roles/jenkins_slave/tasks/main.yml
thrix: Yep, that is a better approach ;) thanks, will post an update

robled:
This PR has been waiting on updates for a while.  Feel free to re-open it if there are any new responses to the review.

PR #74, opened 2017-04-12
initial unit tests and code coverage

robled:
* Added basic unit tests for our CLI to check the exit codes when we
call Ansible

* Added py.test and codecov to travisfile to generate reports at
codecov.io

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,14 @@ addons:
       - debian-sid
     packages:
       - shellcheck
-install: tests/install.sh
-script: tests/${TEST}
+install:
+  - tests/install.sh
+  - pip install pytest-cov
+script:
+  - tests/${TEST}
+  - py.test --cov=cinch/bin --cov-config .coveragerc --cov-report term --cov-report xml --cov-report html tests
+after_success:
+  - pip install codecov && codecov
 
 # Test matrix will be constructed from here. Listing out each set of tests
 # to execute will permit them to be executed in parallel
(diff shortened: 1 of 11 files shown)


.travis.yml
```diff
@@ -14,8 +14,14 @@ addons:
       - debian-sid
     packages:
       - shellcheck
-install: tests/install.sh
-script: tests/${TEST}
+install:
+  - tests/install.sh
+  - pip install pytest-cov
```
greg-hellings: Why not add this to tests/install.sh instead of adding it here? Then we get the benefit of shell linting from shell check, and it can be partnered with the existing pip commands.

.travis.yml
```diff
@@ -14,8 +14,14 @@ addons:
...
+script:
+  - tests/${TEST}
+  - py.test --cov=cinch/bin --cov-config .coveragerc --cov-report term --cov-report xml --cov-report html tests
```
greg-hellings: This will run the coverage test at least 4 times for each commit. Instead, you should add this line to a new shell script under the tests/ directory and then add that shell script to the env matrix.

.travis.yml
```diff
@@ -14,8 +14,14 @@ addons:
...
+after_success:
+  - pip install codecov && codecov
```
greg-hellings: This should probably live inside of the new shell script (with the pip install in the install.sh), as well. Just make sure that you have "set -e" in the script so that this only runs when the tests are successful.

greg-hellings (changes requested):

greg-hellings (approved):

PR #75, opened 2017-04-14
Auto deploy

greg-hellings:
This *should* result in code that auto-deploys to PyPI and Docker Hub when a tag is pushed to GitHub.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -1,0 +1,10 @@
+#!/bin/bash -e
+
+CINCH="$(readlink -f "$(dirname "$0")/../")"
+# Extracts the version line from the setup.py script, and trims off the rest of the line to leave
+# only the expected version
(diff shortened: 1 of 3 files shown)


scripts/deploy.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -e
```
robled, line 1: Wrap this code at 80 characters.

robled (changes requested):

robled (approved):

robled:
This patch is approved, but I'm not merging it myself since it contains release automation bits and it's probably best if the submitter merges it since they have a better understanding of how merging it could affect any triggers or automation that would result.

greg-hellings:
Merging now. Just having the code in the branch /shouldn't/ mess anything up. The deploy code should only activate when a tag is pushed, which I'll do in a second, after this merge.

Issue #67, opened 2017-04-07
Support Running Cinch Container


greg-hellings:
Support a Jenkins Slave configuration that does nothing more than execute the latest Cinch-built Docker container directly on a host.

PR #76, opened 2017-04-17
Add docker slave

greg-hellings:
Add ability to spin up Docker slaves
Factor out wait for calls in Jenkins master to separate file
Add flush handlers and wait for Master to be up at end of role

### DIFF
--- a/vagrant/docker_slave/Vagrantfile
+++ b/vagrant/docker_slave/Vagrantfile
@@ -1,0 +1,17 @@
+require '../shared.rb'
+
+Vagrant.configure("2") do |config|
+	vm(config, "master")
+	vm(config, "docker-slave") do |ansible|
+		ansible.groups = {
+			"jenkins_master" => ["master"],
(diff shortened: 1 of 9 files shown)


greg-hellings:
Addresses #67

vagrant/docker_slave/Vagrantfile
```diff
@@ -0,0 +1,17 @@
+require '../shared.rb'
```
robled, line 1: The indent level on this file is strange to me - is this a ruby thing or can we indent at a standard of four spaces?

robled (changes requested):

vagrant/docker_slave/Vagrantfile
greg-hellings: Nah, this is what code looks like when it's written The Correct Way.

With tabs, instead of spaces. Now excuse me while I go fan some other flames over here...

(PR updated with spaces instead of tabs)

robled (approved):

PR #83, opened 2017-04-19
Update Vagrant with OpenStack

greg-hellings:
Add code needed to run Vagrant against openstack
Update code with some RHEL/OS image based details
Modify Vagrantfile host vars to introspect URLs

### DIFF
--- a/vagrant/shared.rb
+++ b/vagrant/shared.rb
@@ -47,7 +47,7 @@ def vm(config, name, base_box="centos/7")
         end
         nodeconfig.vm.provider :openstack do |os, override|
             os.image = get_image(base_box)
-            os.server_name = ENV['USER'] + name
+            os.server_name = ENV['USER'] + '-' + name
             override.ssh.username = get_username(base_box)
             override.ssh.pty = true
         end
--- a/vagrant/shared.rb
+++ b/vagrant/shared.rb
@@ -2,7 +2,7 @@ ENV['VAGRANT_NO_PARALLEL'] = 'yes'
 $ip = 2
 
 def get_username(base_box)
-	if base_box.start_with?('centos')
+    if base_box.start_with?('centos')
         return 'centos'
     elsif base_box.start_with?('rhel')
         return 'root'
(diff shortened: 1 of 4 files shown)


vagrant/shared.rb
```diff
@@ -1,12 +1,58 @@
 ENV['VAGRANT_NO_PARALLEL'] = 'yes'
 $ip = 2
 
+def get_username(base_box)
+	if base_box.start_with?('centos')
```
tehsmyers: `vagrant/shared.rb|11 warning| mismatched indentations at 'end' with 'if' at 5`

looks like a tab snuck in

vagrant/shared.rb
```diff
@@ -1,12 +1,58 @@
...
+        # Provider-specific setup
+        nodeconfig.vm.provider :libvirt do |lv, override|
+            local_setup(override, base_box)
+        end
+        nodeconfig.vm.provider :virtualbox do |vb, override|
+            local_setup(override, base_box)
+        end
+        nodeconfig.vm.provider :openstack do |os, override|
+            os.image = get_image(base_box)
+            os.server_name = ENV['USER'] + name
```
tehsmyers: this is only aesthetic, but some form of `user + '-' + name` would be nice

vagrant/slave/Vagrantfile
```diff
@@ -9,7 +9,7 @@ Vagrant.configure("2") do |config|
             "cent7" => ["master", "slave"],
             "repositories" => ["master", "slave"],
             "jenkins_slave:vars" => {
-                "jenkins_master_url" => "http://192.168.8.2:8080",
+                "jenkins_master_url" => "http://{{ hostvars['master']['ansible_default_ipv4']['address'] }}:8080",
```
tehsmyers, line 12: nice

tehsmyers (changes requested):
Only thing that needs to be fixed is the tab

tehsmyers (approved):

PR #86, opened 2017-04-22
Ensure ansible-lint compatibility.

jpaulovic:
All lint errors should be fixed.
Mainly indentation fixes, using handlers where appropriate, ditching the use of 'shell' module in favor of 'command' module.

### DIFF
--- a/cinch/group_vars/cent6
+++ b/cinch/group_vars/cent6
@@ -1,20 +1,20 @@
 gcc_compat_package: compat-gcc-34
 
 _repositories:
-  - name: epel-testing
-    mirrorlist: "{{ fedora_mirrors }}repo=testing-epel6"
-  - name: epel
-    mirrorlist: "{{ fedora_mirrors }}repo=epel-6"
-  - name: cloud
-    baseurl: "{{ centos_mirrors }}/$releasever/cloud/$basearch/openstack-juno/"
+- name: epel-testing
+  mirrorlist: "{{ fedora_mirrors }}repo=testing-epel6"
+- name: epel
+  mirrorlist: "{{ fedora_mirrors }}repo=epel-6"
+- name: cloud
+  baseurl: "{{ centos_mirrors }}/$releasever/cloud/$basearch/openstack-juno/"
 
 _download_repositories:
-  - http://pkg.jenkins-ci.org/redhat-stable/jenkins.repo
-  - https://beaker-project.org/yum/beaker-client-CentOS.repo
+- http://pkg.jenkins-ci.org/redhat-stable/jenkins.repo
+- https://beaker-project.org/yum/beaker-client-CentOS.repo
 
 rpm_key_imports:
-  - key: http://pkg.jenkins-ci.org/redhat-stable/jenkins.io.key
-    validate_certs: true
+- key: http://pkg.jenkins-ci.org/redhat-stable/jenkins.io.key
+  validate_certs: true
 
 jenkins_slave_repositories: "{{ _repositories }}"
 jenkins_slave_download_repositories: "{{ _download_repositories }}"
--- a/cinch/install-rhel7.yml
+++ b/cinch/install-rhel7.yml
@@ -7,7 +7,7 @@
 # This playbook was tested with Ansible 1.8.4.
 
 - name: install cinch into a virtualenv on RHEL7
-  hosts: localhost
+  connection: local
   vars:
     jenkins_home: /var/lib/jenkins
     venv_dir: "{{ jenkins_home }}/opt/cinch"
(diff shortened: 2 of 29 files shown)


robled (approved):

cinch/group_vars/cent6
```diff
@@ -1,20 +1,20 @@
 gcc_compat_package: compat-gcc-34
 
 _repositories:
-  - name: epel-testing
-    mirrorlist: "{{ fedora_mirrors }}repo=testing-epel6"
-  - name: epel
-    mirrorlist: "{{ fedora_mirrors }}repo=epel-6"
-  - name: cloud
-    baseurl: "{{ centos_mirrors }}/$releasever/cloud/$basearch/openstack-juno/"
+- name: epel-testing
```
greg-hellings, line 4: I think we need to have a group discussion about using this indentation method for YAML. Personally, I am opposed to it, but I'm willing to have a discussion on it.

cinch/install-rhel7.yml
```diff
@@ -7,7 +7,7 @@
 # This playbook was tested with Ansible 1.8.4.
 
 - name: install cinch into a virtualenv on RHEL7
-  hosts: localhost
+  connection: local
```
greg-hellings, line 10: I'm not sure this change was correct. We don't want to execute this a dozen times locally if someone passes in an inventory script with a dozen different hosts. We really do only want this to execute on localhost.

cinch/roles/jenkins_common/defaults/main.yml
```diff
@@ -1,10 +1,8 @@
-jswarm_version: 2.1
-jenkins_slave_username: ''
-jenkins_slave_password: ''
 # The home directory for the user who will be executing the swarm process
 jenkins_user_home: /var/lib/jenkins
 # The name of the user who will be executing the swarm process
 jenkins_user: jenkins
+jenkins_user_password: changeme
```
greg-hellings, line 5: Why was this added?

cinch/roles/jenkins_master/tasks/iptables.yml
```diff
@@ -40,12 +40,7 @@
     jump: REDIRECT
     to_ports: "{{ https_enabled | ternary(8443, 8080) }}"
     state: present
-  register: real_address_rule
-
-- name: save iptables
-  shell: service iptables save
-  changed_when: false
-  when: forward_rule.changed or loopback_rule.changed or real_address_rule.changed
+  notify: save iptables
```
greg-hellings, line 43: This change breaks iptables. This command needs to be run at precisely this point, before iptables is starts below, or the old rules will be read back into the system. I'll be reverting this change in an upcoming PR.

cinch/roles/jenkins_slave/tasks/main.yml
```diff
@@ -75,15 +70,10 @@
     dest: /etc/systemd/system/swarm.service
     owner: root
     mode: 0644
-  notify: restart swarm
+  notify:
+  - restart swarm
+  - reload systemd files
```
greg-hellings, line 75: This change will break swarm restarts and will need to be reverted, as well.

cinch/install-rhel7.yml
greg-hellings: Indeed, this change actually causes the playbook to still pass Ansible lint, but causes it to fail ansible-playbook --syntax-check (this should probably be something that gets files as an ansible-lint bug!!). I'm preparing a PR that reverts this change and also adds checking this file to our lint.sh script.

cinch/roles/jenkins_common/defaults/main.yml
jpaulovic: Before this change, the jenkins user was being created both in jenkins_common and in jenkins_slave, using the same username (jenkins), with the difference that in jenkins_common you specified more details, but no password and in jenkins_slave you specified just username and password. Since it's the same user, I decided to merge this to create the user in one task here. Maybe I misunderstood and the password should only be set on slave, in that case I'd suggest reverting this and renaming the task from "create jenkins user" to "change password for jenkins user"

cinch/roles/jenkins_master/tasks/iptables.yml
jpaulovic: You are right, my mistake. I'd prefer adding a meta task to flush handlers here and in other roles where necessary, instead of going back to when.changed (which breaks lint rules).
Thanks for pointing this out.

cinch/roles/jenkins_slave/tasks/main.yml
jpaulovic: detto - need meta task for flushing handlers

cinch/install-rhel7.yml
jpaulovic: Fair point, I'd probably go for a combination:
- hosts: 127.0.0.1 //(or localhost)
  connection: local

as suggested in official Ansible docs.

PR #90, opened 2017-04-25
Fedora

greg-hellings:
Updates playbooks to support Fedora
Adds support for Fedora builds to Travis
Adds support for Fedora image deployments to Travis


greg-hellings:
I know it's unorthodox for me to merge this without a review, but I cannot test anything without the changes this makes to the iptables. So I'm going to merge it now and I can update it in the future if there are changes that should be made. 

@robled @seandst

PR #91, opened 2017-04-25
Ansible lint

greg-hellings:
Lint ALL THE THINGS

Unbreak the linting that's there.

PR #92, opened 2017-04-26
Ansible 23 warnings

greg-hellings:
Depends on (and includes the changes from) PR #91 . Do not merge before that.

### DIFF
--- a/tests/ansible_lint_rules/NoFormattingInWhenRule.py
+++ b/tests/ansible_lint_rules/NoFormattingInWhenRule.py
@@ -1,0 +1,32 @@
+from ansiblelint import AnsibleLintRule
+
+
+class NoFormattingInWhenRule(AnsibleLintRule):
+    id = 'CINCH0001'
+    shortdesc = 'No Jinja2 in when'
+    description = '"when" lines should not include Jinja2 variables'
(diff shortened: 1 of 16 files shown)


tests/ansible_lint_rules/NoFormattingInWhenRule.py
```diff
@@ -0,0 +1,32 @@
+from ansiblelint import AnsibleLintRule
```
robled, line 7: Let's add some comments to this file for the functions so that folks new to ansible-lint can dive in more easily.

robled (changes requested):

tests/ansible_lint_rules/NoFormattingInWhenRule.py
greg-hellings: https://github.com/willthames/ansible-lint/blob/master/README.md I just read the docs (and then found out they were wrong, and fixed them by reading the code). This file also lives on, in an updated and improved form, in my PR against upstream. https://github.com/willthames/ansible-lint/pull/262

This file should not last long, if mine gets merged up there.

robled (approved):

robled (approved):

PR #93, opened 2017-04-26
Removing jenkins_plugin.py for upstream version

greg-hellings:
This file has been extraneous for a while. We should remove this and stick with the fixes that are present in the Ansible tree post 2.2.1.


robled:
Does **install_requires** in setup.py need to be updated to ensure we get the proper code from upstream?

greg-hellings:
Negative. We depend on linchpin >= 0.9.0, which has been pinned to ansible==2.2.1. That's the version that included the fixes in this file. However, I'll take this opportunity to update our own direct ansible include to be >=2.2.1 both to bring us in line with linchpin and also to reflect this change directly.

robled (approved):

Issue #31, opened 2017-02-07
Configure Swarm Port


greg-hellings:
Configure Jenkins Master to listen on port 50000 for slaves, instead of relying on the default ports.

greg-hellings:
#101 configures this for masters. Have you tested that slaves can still connect?

jpaulovic:
I have tested the deployment of both master and slave in one playbook run, after which I could see the slave listed in the master under 'Build Executor Status' with a message saying 'Connected via JNLP agent'. I didn't do any other additional checks.

Issue #23, opened 2017-02-07
Firewall updates


greg-hellings:
Ensure the iptables has the following rules:

* Accept all ICMP
* Accept all loopback
* Reject outgoing connections to 127.* that are not on loopback
* Allow inbound SSH
* Accept related and established connections
* Allow inbound TCP port 50000 for Jenkins slaves
* Allow inbound TCP port 8888 for ZeroMQ

Drop/deny everything else.

greg-hellings:
Existing rules should be cleared out and replaced with this.

Additional rules should be added immediately before the drop/deny to do port opening/redirection for 80 and 443 based on HTTPS configuration.

jpaulovic:
I believe all the rules were implemented in #95 , we will remove the port forwarding when dealing with #32 / #88 .

greg-hellings:
Per @jpaulovic's comment and PRs, I'm closing this issue out.

PR #95, opened 2017-04-28
Properly configure iptables

jpaulovic:
Possibly resolving Issue #31 (configure swarm port) - please check
Properly configure iptables - Issue #23


cinch/roles/jenkins_master/tasks/configure.yml
```diff
@@ -23,5 +23,14 @@
     group: jenkins
     state: directory
 
+- name: set jenkins slaveAgentPort
+  template:
+    src: config.xml.j2
```
robled: Something we've found here style-wise is that naming templates with a .j2 extension does follow Ansible/Jinja conventions, but the downside it that it confuses editor syntax highlighting, and we've generally avoided this naming convention as the value of easier editing outweighs the standardization benefits in our view.  I believe our other templates are omitting the .j2 extension.

cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -108,3 +108,11 @@ keystore_pass: changeme
 # like CentOS or RHEL <= 7, this is the yum conf file, but for Fedora and the
 # like, this is the dnf conf file
 version_pin_file: /etc/yum.conf
+# This is the port Jenkins will listen on for slaves.
+jenkins_slave_agent_port: 50000
+# This is a list of ports to allow in firewalld.
```
robled: It looks like we're actually using iptables for this instead of firewalld, so we may want to update this comment to reflect that.

cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -108,3 +108,11 @@ keystore_pass: changeme
...
+firewall_ports:
```
robled: We should indicate that all of these are TCP ports, perhaps by naming ths list **firewall_tcp_ports** or similar.

cinch/roles/jenkins_master/tasks/iptables.yml
```diff
@@ -48,13 +80,19 @@
     state: present
   register: output_port_forward
 
+- name: set iptables to reject all other connections
```
robled, line 89: Is it possible to set this as the default policy for the INPUT chain rather than as an explicit ordered rule?  Additionally, could we send an ICMP message with some further info to inform users for debugging, such as **icmp-admin-prohibited**?

cinch/roles/jenkins_master/tasks/iptables.yml
```diff
@@ -10,6 +10,38 @@
     state: stopped
     enabled: false
 
+- name: set iptables to accept all icmp
```
robled: Can we add comments to these iptables rules so that if someone were inspecting a host configuration they could know the intent of these rules?

cinch/roles/jenkins_master/tasks/iptables.yml
```diff
@@ -10,6 +10,38 @@
...
+    protocol: icmp
+    jump: ACCEPT
+
+- name: set iptables to accept all communication from lo interface
+  iptables:
+    chain: INPUT
+    in_interface: lo
+    jump: ACCEPT
+
+- name: set iptables to reject other communication to localhost
```
robled: Looking at some reference material, this rule may be closer to what we're porting over from our older code:

-A INPUT -d 127.0.0.0/8 ! -i lo -m comment --comment "reject local traffic not on loopback interface" -j REJECT --reject-with icmp-port-unreachable

Perhaps there is a reason to change it, but I'm not aware of a reason at this time.

cinch/roles/jenkins_master/tasks/iptables.yml
```diff
@@ -10,6 +10,38 @@
...
+  iptables:
+    chain: INPUT
+    destination: 127.0.0.1/8
+    jump: REJECT
+
+- name: set iptables to allow established and related connections
```
robled: This rule should be last in the INPUT chain for best performance.

cinch/roles/jenkins_master/tasks/configure.yml
robled: Let's rename this play to indicate that it's doing much more than just setting the slave agent port.

robled (changes requested):

cinch/roles/jenkins_master/templates/config.xml.j2
```diff
@@ -0,0 +1,34 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
robled: Templating this entire config file will require a bit of auditing, in that there are many settings we are configuring here.  Some of these settings may exist in this project as variables ( e.g. <useSecurity>), so we really need to audit the rest of the cinch code to make sure that putting this entire template in place doesn't break other settings.  We also need to be sure later things we're doing don't interfere with this config file and break idempotence.

cinch/roles/jenkins_master/defaults/main.yml
greg-hellings: Alternatively, we can change this from just a list of port numbers to a list of port/protocol numbers.

cinch/roles/jenkins_master/tasks/iptables.yml
greg-hellings: Shouldn't it be the first rule?

cinch/roles/jenkins_master/tasks/iptables.yml
```diff
@@ -48,13 +80,19 @@
...
+    chain: INPUT
+    protocol: tcp
+    jump: REJECT
+
 - name: save iptables
   command: service iptables save
   when: (remote_port_forward|changed) or (local_port_forward|changed) or (output_port_forward|changed)
   tags:
-    - skip_ansible_lint
+  - skip_ansible_lint
```
greg-hellings: Please restore this back to the proper indent.

cinch/roles/jenkins_master/templates/config.xml.j2
```diff
@@ -0,0 +1,34 @@
...
+<hudson>
+  <disabledAdministrativeMonitors/>
+  <version>1.0</version>
+  <numExecutors>10</numExecutors>
+  <mode>NORMAL</mode>
+  <useSecurity>true</useSecurity>
```
greg-hellings: I know this value is something we permit the user to configure.

greg-hellings (changes requested):

cinch/roles/jenkins_master/tasks/iptables.yml
robled: You're correct - I was backwards.

cinch/roles/jenkins_master/tasks/iptables.yml
jpaulovic: Done

jpaulovic:
Can we merge this now?

cinch/roles/jenkins_master/tasks/iptables.yml
```diff
@@ -10,6 +10,42 @@
...
+    comment: "001 accept all to lo interface"
+
+- name: set iptables to reject other communication to localhost
+  iptables:
+    chain: INPUT
+    destination: 127.0.0.1/8
+    reject_with: icmp-port-unreachable
+    comment: "002 reject local traffic not on loopback interface"
+
+- name: set iptables to allow established and related connections
```
robled: Let's set this as the very first rule in the chain for best performance.

robled (changes requested):

cinch/roles/jenkins_master/tasks/iptables.yml
jpaulovic: Done.
Forgot about this, sorry...

robled (approved):

greg-hellings (approved):

PR #98, opened 2017-05-02
Issue45

greg-hellings:
Brings code in line with ansible-lint and fixes several issues with SSL masters.


tests/ansible_lint_rules/NoFormattingInWhenRule.py
```diff
@@ -1,20 +1,19 @@
-# Documentation on AnsibleLintRules can be found here:
```
robled, line 1: Did these comments need to be removed?

robled (changes requested):

tests/ansible_lint_rules/NoFormattingInWhenRule.py
robled: I assume that the file got upstreamed, so perhaps that's why the comments are gone.  Just verifying.

greg-hellings: I'm trying to keep this file in sync with the PR upstream. Will had requested some changes, mainly around Python 3 compatibility, so I pulled them in here when I added the until code, so I could push the until code back into that PR.

Basically it's just a bunch of back and forth until it reaches its final form.

robled (approved):

PR #102, opened 2017-05-03
Issue18

greg-hellings:
Bring configuration ability in line with Jenkins Puppet config options.


jpaulovic:
Hi, I just bumped into an undefined variable today in the jenkins master role, specifically {{ ansible_user_dir }} used in cli-users.yml and plugins.yml. It doesn't seem to break the execution, but I find it kind of strange that it's not defined anywhere and yet used. Is it a bug? It seemed appropriate to bring it up here when I saw you were dealing with the jenkins_master role's defaults.

greg-hellings:
ansible_user_dir is a variable that is native to Ansible and should be set equal to the $HOME value of the currently connected user. It should definitely not be defined in our code anywhere.

robled (approved):

Issue #22, opened 2017-02-07
SSH


greg-hellings:
Ensure that SSH is installed and running on Jenkins masters.

PR #103, opened 2017-05-04
Introduce the option to add ssh keys to authorized_keys

jpaulovic:
Possibly related to #22 .


cinch/roles/jenkins_common/tasks/main.yml
```diff
@@ -43,3 +43,10 @@
     owner: "{{ jenkins_user }}"
     mode: 0600
   notify: generate public SSH key
+
+- name: add ssh keys to authorized_keys
```
robled: Let's add "for the jenkins user" to the name of this task so in log output it's a bit more clear what this is doing.

robled (changes requested):

cinch/roles/jenkins_common/defaults/main.yml
```diff
@@ -7,3 +7,5 @@ jenkins_ssh_keyfile: id_rsa
 java_version: 8
 # Needed for certain OpenShift instances
 jenkins_user_uid: 1000090000
+# Set the default (empty) set of authorized keys for jenkins user.
```
greg-hellings: You might want to add a note that this should be an array of file names on the local path pointing to public keys, so someone doesn't try to add the key values directly into this array or blindly add private keys or some other such nonsense.

cinch/roles/jenkins_common/defaults/main.yml
```diff
@@ -7,3 +7,8 @@ jenkins_ssh_keyfile: id_rsa
 java_version: 8
 # Needed for certain OpenShift instances
 jenkins_user_uid: 1000090000
+# Set the default (empty) set of authorized SSH public keys for jenkins user.
+# To add authorized keys, override this value by adding a list of filenames
+# in the local path pointing to separate public keys.
```
robled: Can the given path be relative or is a full path required?  I find that sometimes Ansible needs a full path in certain environments, and we should always document whether a relative or full path is acceptable.

jpaulovic: Both relative and full paths work. I updated the wording to state that.

robled (approved):

cinch/roles/jenkins_common/defaults/main.yml
jpaulovic: Documentation updated.

Issue #99, opened 2017-05-02
(1) Add ability to pin plugins during plugin install


scoheb:
In some cases, we want to pin a plugin to a specific version. The process is explained here: https://wiki.jenkins-ci.org/display/JENKINS/Pinned+Plugins . (Note: this is only applicable to Jenkins 1.X)

Given a list of plugin names
When a jenkins instance is deployed
Then each plugin name will have a file created in JENKINS_HOME/plugins/{plugin_name}.jpi.pinned

Example:
pinned_plugins=credentials,junit

will create:
- JENKINS_HOME/plugins/credentials.jpi.pinned
- JENKINS_HOME/plugins/junit.jpi.pinned

PR #105, opened 2017-05-04
Provide mechanism to pin plugins

greg-hellings:
Provides a method to pin arbitrary Jenkins plugins

Addresses issue #99

### DIFF
--- a/cinch/roles/jenkins_master/defaults/main.yml
+++ b/cinch/roles/jenkins_master/defaults/main.yml
@@ -32,6 +32,10 @@ jenkins_version: "1.651.3"
 # The list of Jenkins plugins that should be installed by default from the
 # configured Update Centers.
 jenkins_plugins: "{{ lookup('file', 'files/jenkins-plugin-lists/default.txt').split('\n') }}"
+# A list of Jenkins plugins that should not be upgraded, but should, instead, be
+# pinned to the currently available versions. Make this an array of the names of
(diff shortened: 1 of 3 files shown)


greg-hellings:
@scoheb Please also check this out

robled (approved):

cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -32,6 +32,10 @@ jenkins_version: "1.651.3"
 # The list of Jenkins plugins that should be installed by default from the
 # configured Update Centers.
 jenkins_plugins: "{{ lookup('file', 'files/jenkins-plugin-lists/default.txt').split('\n') }}"
+# A list of Jenkins plugins that should not be upgraded, but should, instead, be
```
scoheb: I would change it to be:

> Bundled plugins are ones that are included by default with Jenkins. The process of pinning a bundled plugin tells Jenkins to not overwrite the correct version of plugin with an included version during a Jenkins upgrade but rather use the one in JENKINS_HOME. Make this an array of the names of the plugins that should be pinned. NOTE: This only works for Jenkins 1.X

scoheb:
Where do we mention the order of when pinning should happen?

greg-hellings:
We don't need to mention it, we just put it at the correct location in the code. I've included it before the installation of the plugins, as you indicated.

Issue #72, opened 2017-04-11
Catch systemd/upstart service failures for swarm


robled:
I've witnessed a case where the systemd swarm service started via Ansible, but later failed due to an authentication issue with the Jenkins master.  In these cases we should watch the swarm service for some period of time to make sure it started successfully.

greg-hellings:
@robled Can this be closed now that #107 has been merged?

PR #107, opened 2017-05-08
Added error reporting for swarm service failures via systemd and upstart #72

robled:
#72 

* Modified service templates to allow the init system to manage swarm's
retry behavior

* Ansible will now fail if the swarm client cannot connect to the
master, and will pretty-print the error message for human consumption.

* Service timeouts and retry delays have been set to 10 seconds to align
with the default behavior of the swarm client.

* Re-ordered lines in service templates to make them easier to compare
side-by-side

* Fixed an issue in the upstart service template where the
authentication parameters were not set

### DIFF
--- a/cinch/roles/jenkins_slave/tasks/check_swarm_systemd.yml
+++ b/cinch/roles/jenkins_slave/tasks/check_swarm_systemd.yml
@@ -1,0 +1,17 @@
+- name: check if swarm connected to the master with systemd
+  command: systemctl status swarm
+  ignore_errors: yes
+  changed_when: false
+  register: service_swarm_status
+
+- name: grab errors for the last run of the systemd swarm service
+  shell: >
+    journalctl --no-pager -u swarm -o cat
+    _PID=$(systemctl show swarm --property=ExecMainPID
+    | cut -d '=' -f 2)
+  register: service_swarm_err_output
+  when: service_swarm_status|failed
+
+- name: report errors for systemd swarm service
+  include: check_swarm_errors.yml
+  when: service_swarm_status|failed
--- a/cinch/roles/jenkins_slave/tasks/check_swarm_systemd.yml
+++ b/cinch/roles/jenkins_slave/tasks/check_swarm_systemd.yml
@@ -1,9 +1,14 @@
 - name: check if swarm connected to the master with systemd
   command: systemctl status swarm
-  ignore_errors: yes
+  ignore_errors: true
   changed_when: false
   register: service_swarm_status
 
+# Some shell magic is required here to grab the log output from *only* the last
+# run of the swarm service in systemd.  To do this, we find the last PID that
+# was used to start the service using systemctl, and we extract just the PID
+# via 'cut'.  Then we pass that PID to journalctl to give us only the log data
+# we want.
 - name: grab errors for the last run of the systemd swarm service
   shell: >
     journalctl --no-pager -u swarm -o cat
--- a/cinch/roles/jenkins_slave/tasks/check_swarm_upstart.yml
+++ b/cinch/roles/jenkins_slave/tasks/check_swarm_upstart.yml
@@ -1,0 +1,14 @@
+- name: check if swarm connected to the master with upstart
+  command: "grep 'INFO: Connected' /var/log/jenkins-swarm.log"
+  ignore_errors: yes
+  changed_when: false
+  register: service_swarm_status
+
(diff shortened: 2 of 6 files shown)


tehsmyers (approved):
LGTM, but my ansible-fu isn't strong enough to say if there's a better option than the conditional includes, so another ack might be good. The comments make some of the eyebrow-raisers make sense, so thanks for those.

tehsmyers:
As far as I can tell, the travis failure is environmental, related to f25 having trouble updating with yum.

cinch/roles/jenkins_slave/templates/swarm.service
```diff
@@ -11,10 +24,10 @@ ExecStart={{ jenkins_java_cmd }} \
           -executors "$SWARM_EXECUTORS" \
           -labels "$SWARM_SLAVE_LABEL" \
           -fsroot "$SWARM_ROOT" \
+          -retry 0 \
```
jpaulovic: are you sure you want to limit it so strictly? Wouldn't at least one retry make sense here?

cinch/roles/jenkins_slave/tasks/check_swarm_systemd.yml
```diff
@@ -0,0 +1,17 @@
+- name: check if swarm connected to the master with systemd
+  command: systemctl status swarm
+  ignore_errors: yes
+  changed_when: false
+  register: service_swarm_status
+
+- name: grab errors for the last run of the systemd swarm service
+  shell: >
```
greg-hellings: Mmm, shell magic. <3

cinch/roles/jenkins_slave/tasks/check_swarm_systemd.yml
```diff
@@ -0,0 +1,17 @@
+- name: check if swarm connected to the master with systemd
+  command: systemctl status swarm
+  ignore_errors: yes
```
greg-hellings: Can we go with "true" here, instead of "yes"? The rest of the codebase sticks with true/false instead of yes/no.

cinch/roles/jenkins_slave/tasks/check_swarm_upstart.yml
```diff
@@ -0,0 +1,14 @@
+- name: check if swarm connected to the master with upstart
```
greg-hellings, line 1: Does upstart not provide us the niceties that systemd offers?

cinch/roles/jenkins_slave/templates/swarm.service
```diff
@@ -2,7 +2,20 @@
 After=network-online.target
 Wants=network-online.target
 [Service]
+User={{ jenkins_user }}
 EnvironmentFile=/etc/sysconfig/jenkins_swarm
+Restart=on-failure
+
+# Wait 10 seconds (swarm client default) for the service to fail; code in
```
greg-hellings: If code elsewhere depends on this, and we want it to match in these two locations, should we pull it out to a single variable so we don't rely on a magic value?

greg-hellings (changes requested):

cinch/roles/jenkins_slave/tasks/check_swarm_systemd.yml
robled: I missed some consistency here.  Will fix.

cinch/roles/jenkins_slave/tasks/check_swarm_systemd.yml
robled: After reading some systemctl and journalctl docs it was the only way I could find to grab the *last* log output of the swarm service.

cinch/roles/jenkins_slave/tasks/check_swarm_upstart.yml
robled: Based on my research it does not, unfortunately.  systemd is way more feature-rich than upstart.  upstart's log/status functionality leaves much to be desired, and if you run 'initctl status swarm' you'll get a zero exit code back even though the state is anything but *OK*, and parsing the output of 'initctl' was not reliable based on my testing.

cinch/roles/jenkins_slave/templates/swarm.service
robled: Good catch, I'll see about getting this added.

greg-hellings (approved):

PR #108, opened 2017-05-09
Setting up nginx SSL termination instead of leaving handling of SSL…

### LINKED ISSUE #32: Replace Apache
Replace Apache SSL termination with Nginx for SSL termination.

jpaulovic:
… to Jenkins.
Should resolve #32


cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -13,18 +13,13 @@
 update_centers:
   - id: default
     url: "https://updates.jenkins-ci.org/{{ jenkins_version }}/update-center.json"
-# If set to true, Jenkins will be run over HTTPS (Jenkins on port 8443 with a
-# firewall forward from port 443). If you set this value to true, you MUST
-# override jenkins_ssl_cert and jenkins_ssl_key in a production system or you
-# will not see any improvements in security
-https_enabled: false
```
greg-hellings: Why have you completely gotten rid of this variable? SSL should still be optional.

cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -152,5 +145,5 @@ firewall_tcp_ports:
   - "{{ jenkins_slave_agent_port }}"
   - "8888"
   - "22"
-  - "{{ https_enabled | ternary('443', '80') }}"
-  - "{{ https_enabled | ternary('8443', '8080') }}"
+  - "80"
+  - "443"
```
greg-hellings: This port should only be opened if SSL is enabled.

cinch/roles/jenkins_master/templates/etc/nginx/conf.d/jenkins.conf
```diff
@@ -0,0 +1,66 @@
...
+  listen 80;
+  server_name {{ inventory_hostname }};
+  return 301 https://$host$request_uri;
+}
+
+server {
+  listen 443 ssl;
+  server_name {{ inventory_hostname }};
+  gzip on;
+    ssl_certificate {{ (jenkins_ssl_cert is defined) | ternary('/var/lib/jenkins/.ssl/{{ inventory_hostname }}.crt', '/etc/nginx/conf.d/ssl.pem') }};
```
greg-hellings: Why the change in indent?

cinch/roles/jenkins_master/templates/etc/nginx/conf.d/jenkins.conf
```diff
@@ -0,0 +1,66 @@
+server_names_hash_bucket_size 6400;
+upstream jenkins {
+  server 127.0.0.1:8080 fail_timeout=0;
+}
+
+server {
```
greg-hellings: This should be able to handle both with and without HTTPS

cinch/roles/jenkins_slave/handlers/main.yml
```diff
@@ -8,3 +8,6 @@
   systemd:
     daemon_reload: true
     name: swarm
+
```
greg-hellings: Why is this here?

cinch/roles/nginx/defaults/main.yml
```diff
@@ -0,0 +1,23 @@
+nginx_capable: true
+httpd_capable: true
+## set some defaults with the expectation that they will be set in/from calling role
+service_name: "{{ inventory_hostname }}"
+
+## nginx core configuration defaults
+nginx_default_port: 80
```
greg-hellings: This doesn't seem to be used in the config file template.

cinch/roles/nginx/files/etc/sysconfig/iptables.d/httpd
```diff
@@ -0,0 +1,2 @@
+-A INPUT -p tcp --destination-port 80 -j ACCEPT
```
greg-hellings: Shouldn't this work be handled by our iptables.yml in the jenkins_master role?

cinch/roles/nginx/templates/etc/nginx/conf.d/errors.conf.snippet.j2
```diff
@@ -0,0 +1,11 @@
+# redirect server error pages to the static pages from ESO_error_pages RPM
```
greg-hellings: In the cinch repository, we have been avoiding the use of the .j2 extension in templates as it breaks most editor highlighting.

greg-hellings (changes requested):

cinch/roles/jenkins_master/defaults/main.yml
jpaulovic: Because, honestly, I don't really see the point of having https disabled if we can automatically generate a self-signed SSL cert when none provided and it would require having 2 separate nginx configuration templates. However if it's really needed I can add it. Let's discuss it tomorrow.

jpaulovic: Re-added in the latest commit.

cinch/roles/jenkins_master/defaults/main.yml
jpaulovic: fixed

cinch/roles/jenkins_slave/handlers/main.yml
jpaulovic: todo: delete

jpaulovic:
All discussed errors should be now fixed.

cinch/roles/nginx/defaults/main.yml
jpaulovic: It's used in the etc/nginx/nginx.conf default configuration

cinch/roles/nginx/templates/etc/nginx/conf.d/errors.conf.snippet.j2
jpaulovic: changed

cinch/roles/jenkins_master/tasks/configure.yml
```diff
@@ -23,5 +23,18 @@
     group: jenkins
     state: directory
 
+- name: set hostname to match inventory hostname
+  hostname:
+    name: "{{ inventory_hostname }}"
+  when: jenkins_ssl_cert is not defined
```
greg-hellings: I'm confused what this task is doing and why. "inventory_hostname" is a completely user-friendly value and shouldn't be relied on for anything technical if this is being used for things like Host: header matching. Can you explain what this is for?

cinch/roles/jenkins_master/tasks/ensure_up.yml
```diff
@@ -1,16 +1,6 @@
...
-
-- name: check for running in a container
-  set_fact:
-    _listening_port: "{{ https_enabled | ternary(8443, 8080) }}"
-  when: ansible_connection == 'docker'
-
 - name: construct Jenkins URL
   set_fact:
-    _jenkins_url: "http{{ https_enabled | ternary('s', '') }}://localhost:{{ _listening_port }}"
+    _jenkins_url: "http{{ https_enabled | ternary('s', '') }}://{{ inventory_hostname }}"
```
greg-hellings: Why have you made these changes to this file? You've removed logic that was necessary to get Docker working properly, and you've changed to using "inventory_hostname" instead of localhost, which can easily lead to an invalid URL as inventory_hostname can be set to something like, "my_computer_of_awesomeness".

cinch/roles/jenkins_master/templates/etc/nginx/conf.d/jenkins_http.conf
```diff
@@ -0,0 +1,46 @@
+server_names_hash_bucket_size 6400;
+upstream jenkins {
+  server 127.0.0.1:8080 fail_timeout=0;
+}
+
+server {
+  listen 80; 
+  server_name {{ inventory_hostname }};
```
greg-hellings: You shouldn't rely on this value to match anything in the real world.

cinch/roles/jenkins_master/templates/etc/nginx/conf.d/jenkins_https.conf
```diff
@@ -0,0 +1,65 @@
+server_names_hash_bucket_size 6400;
+upstream jenkins {
+  server 127.0.0.1:8080 fail_timeout=0;
+}
+
+server {
+  listen 80;
+  server_name {{ inventory_hostname }};
```
greg-hellings: Same here.

cinch/roles/jenkins_master/templates/etc/nginx/conf.d/jenkins_https.conf
```diff
@@ -0,0 +1,65 @@
...
+  return 301 https://$host$request_uri;
+}
+
+server {
+  listen 443 ssl;
+  server_name {{ inventory_hostname }};
```
greg-hellings: And here.

greg-hellings:
If we're now generating a new, unique SSL key/cert combo whenever one is not provided, can you remove the example pair from the repo in this PR?

cinch/roles/jenkins_master/tasks/configure.yml
jpaulovic: Sure. The point is to avoid the error message produced when the hostname of the machine doesn't match the certificate. In that case, the 'ensure_up' task will timeout because Python will fail to verify the SSL cert and therefore successfully connect to the host. This should be handled externally when NOT using the self-signed certificate.

cinch/roles/jenkins_master/tasks/ensure_up.yml
jpaulovic: Fair point, reverting for the sake of Docker.

cinch/roles/jenkins_master/templates/etc/nginx/conf.d/jenkins_http.conf
jpaulovic: Here, I disagree. I can't think about any better way of getting the host's hostname/fqdn. And if you're managing your infrastructure correctly, one would assume that you also use correct hostnames to connect to your hosts via Ansible (or anything else for that matter) - whether it's a real hostname or just an IP address. I currently don't see a better way to get this value considering we need either a valid DNS hostname or an IP address as a fallback.

greg-hellings: Well some examples of a place we don't use either DNS hostname or IP address would be:

* Our Vagrant files use a hostname equal to the Vagrantfile name and set ansible_host
* Our Docker images that get an inventory_hostname equal to their container name
* Machines generated by linchpin which get their topology names configured as inventory hostname and an appropriate value configured to ansible_host
* Anyone who wants to shortcut group names by setting the inventory_hostname to something like "jenkins_master" and then set ansible_host to an fqdn or IP address
* Someone who is configuring behind a firewall or load balancer where hosts have internal-only names and not names that can be reached from the Ineternet
* Basically every single Ansible inventory I've ever worked with anywhere.

You should factor this out to a single value that a user can set but which draws a default value from some set of either the Ansible facts (you can discover the default IP address for the system - see the vagrant directory for some examples of where I do this) or from the inventory but default to the value of ansible_host if it is set (see the check_ssh role for an example of how to do that). But just blindly using inventory_hostname is going to break things.

cinch/roles/jenkins_master/tasks/configure.yml
greg-hellings: So inventory_hostname is not the appropriate value to set here, as that name does not necessarily have to correspond to a value that is even a valid hostname. The better solution is to modify the ensure_up file to work with potentially invalid certificates. Preferably by setting a value like "validate_certs: false" on its uri call.

cinch/site.yml
```diff
@@ -35,6 +35,7 @@
   hosts: jenkins_master
   roles:
     - beaker-client
+    - nginx
```
greg-hellings: This should be guarded with a when: ansible_connection != 'docker' clause, as we don't want extra services running within a Docker context.

jpaulovic:
About the SSL key/cert combo - I don't see any SSL keypair in this repo, only SSH. which seems unrelated. Or am I missing something?

greg-hellings:
It should be in the Vagrant directory

jpaulovic:
@greg-hellings SSL keypair removed. Any other issues?

cinch/roles/jenkins_master/tasks/ensure_up.yml
```diff
@@ -18,6 +18,7 @@
     url: "{{ _jenkins_url }}"
     status_code: 200
     timeout: 5
+    validate_certs: no
```
greg-hellings: Please use "false" instead of "no" for booleans.

cinch/roles/jenkins_master/tasks/https.yml
```diff
@@ -15,34 +15,20 @@
     mode: 0755
   with_items:
     - in: "{{ jenkins_ssl_cert }}"
-      out: jenkins.crt
+      out: "jenkins.crt"
     - in: "{{ jenkins_ssl_key }}"
-      out: jenkins.key
+      out: "jenkins.key"
```
greg-hellings: Why did you change this line and the one before it?

cinch/roles/jenkins_master/tasks/plugins.yml
```diff
@@ -56,7 +56,7 @@
 
 - name: perform plugin install
   jenkins_plugin:
-    url: "{{ _jenkins_url }}"
+    url: "http://localhost:8080"
```
greg-hellings: Why did you change this?

cinch/roles/nginx/defaults/main.yml
```diff
@@ -0,0 +1,22 @@
+nginx_capable: true
+httpd_capable: true
+## set some defaults with the expectation that they will be set in/from calling role
+service_name: "{{ inventory_hostname }}"
```
greg-hellings: This should match the definition back in the master role.

greg-hellings: Alternatively, they should use the same value placed in group_vars/all

cinch/roles/nginx/handlers/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: reload systemd
+  command: systemctl daemon-reload
+  when: ansible_distribution == "RedHat" and ansible_distribution_major_version == "7"
```
greg-hellings: Shouldn't this rely on the value of ansible_service_mgr instead of sniffing out versions in particular?

cinch/roles/nginx/tasks/nginx.yml
```diff
@@ -0,0 +1,67 @@
+- name: install nginx
+  yum:
```
greg-hellings: People use "package" instead of "yum" as it will continue our support for Fedora.

cinch/roles/nginx/tasks/nginx.yml
```diff
@@ -0,0 +1,67 @@
...
+    name: nginx
+    state: installed
+  when: "repos_add is not defined or 'epel' not in repos_add"
+  notify:
+  - reload systemd
+  - restart nginx service
+
+- name: install nginx
+  yum:
```
greg-hellings: Likewise here and throughout this file.

cinch/roles/nginx/tasks/selinux.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: enable selinux overrides
+  seboolean:
+    name: "{{ item }}"
+    state: yes
```
greg-hellings: "yes" Is used here whereas "true/false" is used throughout the rest of this role and Cinch in general.

cinch/roles/nginx/templates/etc/nginx/nginx.conf
```diff
@@ -0,0 +1,40 @@
+user  nginx;
+worker_processes  {{ nginx_worker_processes }};
+
+error_log  /var/log/nginx/error.log {{ nginx_error_level }};
+{% if ansible_distribution_major_version == "7" %}
```
greg-hellings: Does this support Fedora?

greg-hellings:
@jpaulovic It looks like you removed references to the SSL keys from the Vagrantfiles but not the files themselves. Please 'git rm' the files themselves as well.

greg-hellings:
This PR generates a large number of lint errors. Please clean those up, as well as look into why the cent7_master.sh script is failing in Travis CI.

cinch/roles/nginx/tasks/selinux.yml
jpaulovic: This is by design to achieve compliance with official docs of the module: http://docs.ansible.com/ansible/seboolean_module.html
Using true/false may result in undefined behavior now, or in the future. I would really rather stick to officially supported and documented values.

jpaulovic:
Also about: 
[ANSIBLE0006] service used in place of service module
/home/travis/build/RedHatQE/cinch/cinch/roles/jenkins_master/handlers/main.yml:14
Task/Handler: save iptables

-> not much I can do about this since iptables module doesn't support saving iptables config and it was already like this before. I would strongly suggest ditching iptables and moving to firewalld.

[ANSIBLE0019] No Jinja2 in when/until
/home/travis/build/RedHatQE/cinch/cinch/roles/jenkins_master/tasks/iptables.yml:45
Task/Handler: set iptables to allow Jenkins-related TCP ports

-> this is here to ensure the https port is only opened when https is used. It seemed like a nice way to tell Ansible to only open the port when needed in one task and having all the ports stored in one list instead of making a separate variable which would only hold the port 443 if HTTPS was enabled.

cinch/roles/jenkins_master/tasks/plugins.yml
jpaulovic: Because of problems with SSL verification when using self-signed SSL certiifcate. This task would fail as the SSL cert could not be verified.

cinch/roles/jenkins_master/tasks/https.yml
jpaulovic: Strings should always be quoted IMHO. It helps preventing errors when vars contain weird characters and stuff.

greg-hellings: In YAML strings only need to be quoted if they are ambiguous - e.g. if they start with a quotation mark or single tick that needs to be part of the string, if it starts with a YAML control character such as '{' or '[', etc. When a string literal does not begin with any of those, then the quotation marks can be left out like we do in all the other locations.

cinch/roles/jenkins_master/tasks/plugins.yml
greg-hellings: If _jenkins_url is being generated improperly inside of ensure_up.yml, you should change it there instead of skipping over it here. If SSL is your only problem, then you can set validate_certs: false here.

greg-hellings:
The two lint errors: ANSIBLE0006 you can skip linting on that particular task by adding the tag skip_ansible_lint.

ANSIBLE0019: It's complaining about this line - "{{ item }} is defined". Including the {{ }} syntax in a when/until line is deprecated in Ansible. The proper way for this would just be ``when: item is defined``

cinch/roles/nginx/tasks/selinux.yml
greg-hellings: Use of yes/no in place of true/false is a feature of YAML 1.1 that is not available in 1.2. It doesn't have anything in particular to do with Ansible. Throughout the rest of the playbooks we use true/false in order to avoid these YAML 1.1/1.2 inconsistencies.

cinch/roles/nginx/tasks/main.yml
```diff
@@ -16,11 +15,3 @@
   service:
     name: nginx
     enabled: yes
```
greg-hellings: This also falls into the true/false problem.

cinch/roles/jenkins_master/tasks/plugins.yml
jpaulovic: Cool, this option is not listed in the official docs, so I assumed it wouldn't work. But it does! Reverted.

cinch/roles/jenkins_master/templates/sysconfig_jenkins
```diff
@@ -79,7 +79,7 @@ JENKINS_HTTPS_PORT="{{ https_enabled | ternary(8443, '') }}"
 # Path to the keystore in JKS format (as created by the JDK 'keytool').
 # Default is disabled.
 #
-JENKINS_HTTPS_KEYSTORE="{{ jenkins_https_keystore }}"
+JENKINS_HTTPS_KEYSTORE=""
```
greg-hellings, line 82: This variable, and the one below it, which are no longer in use, should be removed from the defaults/main.yml file.

cinch/roles/jenkins_master/tasks/plugins.yml
greg-hellings: I don't see the reversion here.

jpaulovic: fixed

cinch/roles/jenkins_master/templates/sysconfig_jenkins
jpaulovic: done

greg-hellings (approved):

cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -13,18 +13,17 @@
...
-# will not see any improvements in security
+# If set to true, Jenkins will be run over HTTPS. If you set this value to true,
+# you can override jenkins_ssl_cert and jenkins_ssl_key if you want to use custom
+# SSL certificate, otherwise a self signed certificate will be used.
 https_enabled: false
-# Local path to the SSL certificate to use in configuring HTTPS. This MUST be
-# overridden if you want secure access to Jenkins, as these default files are
-# available to the public, thus compromising the security of using these values
-jenkins_ssl_cert: jenkins.crt
+# Local path to the SSL certificate to use in configuring HTTPS. This can be
```
robled, line 20: Can this path be relative?  And if so, relative to what?  To me, this info seems useful to those reading the comment here.

cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -13,18 +13,17 @@
...
+# overridden if you want secure access to Jenkins using your own SSL certificate.
+# By default, a self signed certificate created by the nginx role will be used.
+#jenkins_ssl_cert:
 # Local path to the SSL encryption key to use in configuring HTTPS. See the
```
robled, line 24: Can this path be relative?  And if so, relative to what?  To me, this info seems useful to those reading the comment here.

cinch/roles/jenkins_master/tasks/configure.yml
```diff
@@ -23,5 +23,14 @@
     group: jenkins
     state: directory
 
+- name: install nginx configuration file for jenkins
+  template:
```
robled, line 27: Might be a good idea to explicitly set the mode for this file.

cinch/roles/jenkins_master/tasks/https.yml
```diff
@@ -18,31 +18,18 @@
       out: jenkins.crt
     - in: "{{ jenkins_ssl_key }}"
       out: jenkins.key
+  when: jenkins_ssl_cert is defined and jenkins_ssl_key is defined
   notify: restart Jenkins
 
 - name: copy ssl certificate to anchors
   copy:
-    src: "{{ jenkins_ssl_cert }}"
+    src: "{{ (jenkins_ssl_cert is defined) | ternary('{{ jenkins_home }}/.ssl/jenkins.crt',
```
robled, line 26: Now that keytool is out of the picture, is there any need to store certificates in **$JENKINS_HOME/.ssl**?  I always have seen that path used in conjunction with keytool, and to me it seems that we can just always store certificates where nginx expects to read them by default.

cinch/roles/jenkins_master/templates/etc/nginx/conf.d/jenkins_https.conf
```diff
@@ -0,0 +1,65 @@
...
+  listen 80;
+  server_name {{ service_name }};
+  return 301 https://$host$request_uri;
+}
+
+server {
+  listen 443 ssl;
+  server_name {{ service_name }};
+  gzip on;
+    ssl_certificate {{ (jenkins_ssl_cert is defined) | ternary('/var/lib/jenkins/.ssl/jenkins.crt', '/etc/nginx/conf.d/ssl.pem') }};
```
robled, line 16: Now that keytool is out of the picture, is there any need to store certificates in $JENKINS_HOME/.ssl? I always have seen that path used in conjunction with keytool, and to me it seems that we can just always store certificates where nginx expects to read them by default.

cinch/roles/nginx/files/etc/nginx/conf.d/default.conf
```diff
@@ -0,0 +1,44 @@
+#server {
```
robled, line 1: Why this is file commented out?  If it needs to be this way by default, perhaps add a comment at the top explaining why.

cinch/roles/nginx/handlers/main.yml
```diff
@@ -0,0 +1,8 @@
+- name: reload systemd
```
robled, line 1: Can the Ansible systemd module be used for this?

cinch/roles/nginx/tasks/ssl-setup.yml
```diff
@@ -0,0 +1,49 @@
+- name: copy over ssl key
+  copy:
```
robled, line 2: Should we specify explicit owner/group/mode here?

cinch/roles/nginx/tasks/ssl-setup.yml
```diff
@@ -0,0 +1,49 @@
...
+        - "{{ httpd_ssl_key_file }}"
+      skip: True
+  register: setup_ssl_key
+  notify: restart nginx service
+  no_log: True
+  tags:
+    - update_ssl_certs
+
+- name: copy over ssl pem file
+  copy:
```
robled, line 16: Should we specify explicit owner/group/mode here?

cinch/roles/nginx/templates/etc/nginx/conf.d/monitor.conf.snippet
```diff
@@ -0,0 +1,8 @@
+        location /nginx_status {
+                stub_status on;
+                access_log   off;
+                allow 127.0.0.1;
+		allow 10.8.92.8;
```
robled, line 5: Should these class A IP addresses be included in this file by default?

cinch/roles/nginx/handlers/main.yml
jpaulovic: only in Ansible 2.4 and above (at least that's what I understood from the docs). I'll test this out... I guess we claim compatibility also with lower versions, mainly since 2.4 is not yet available even on Fedora...

greg-hellings: 2.4 hasn't even been released yet. However, that is the next version that will come out, so that probably means the current git master of Ansible has support for this.

cinch/roles/nginx/templates/etc/nginx/conf.d/monitor.conf.snippet
jpaulovic: certainly not

cinch/roles/nginx/tasks/ssl-setup.yml
jpaulovic: hmm, good question. I tried to modify the existing role as little as possible, so I'm not exactly sure... 
It isn't really used anyway, since the SSL cert which is really used gets copied in roles/jenkins_master/tasks/https.yml
I'll try to simplify the role as much as possible to remove all this unnecessary stuff since we agreed to incorporate external roles in later phases...

cinch/roles/nginx/handlers/main.yml
robled: Reading the Ansible docs on this feature indicates (at least to me) that it's supported in 2.2:

https://docs.ansible.com/ansible/systemd_module.html

Am I missing something there?  Now there is a note at the bottom about special conditions for 2.4, but those appear to be some required options, but I don't think that prevents us from using the module in 2.2.

jpaulovic: yeah, you're actually right. I've changed it in the new PR #116 and it seems to work...

PR #112, opened 2017-05-12
Don't use cached pip when building in travis

tehsmyers:
This is a test PR. Since travis is failing on f25 in the "setting up
build cache" step, I'm just taking a stab at seeing if it contiues to
break if we don't use the cache.


tehsmyers:
Totally failed.

PR #115, opened 2017-05-22
Added idempotent management of Jenkins config.xml settings

robled:
* Moved groovy and cli-users tasks to a new file that will be executed
towards the end of the role where all necessary dependencies have been
installed so that we can manage Jenkins application settings.

* Jenkins global config settings changes are now idempotent and we do
not restart Jenkins when making these changes to retain uptime.

* A single task has been added to execute any number of Groovy scripts
in a standard fashion and in the specified order.


cinch/roles/jenkins_master/templates/set_slaveport.groovy
```diff
@@ -0,0 +1,14 @@
+import jenkins.model.Jenkins
+def jenkins = Jenkins.instance
+
+def slave_agent_port = jenkins.getSlaveAgentPort()
+int jenkins_slave_agent_port = {{ jenkins_slave_agent_port|int }}
+
+if (slave_agent_port != jenkins_slave_agent_port) {
+    jenkins.setSlaveAgentPort(jenkins_slave_agent_port)
+    jenkins.save()
+    println "CHANGED: slave agent port from " + slave_agent_port + " to " + \
```
greg-hellings, line 10: I'm not sure the continuation character is necessary, as Groovy is really just Java. But it also shouldn't hurt anything.

greg-hellings (approved):

PR #116, opened 2017-05-23
Move SSL configuration entirely to the nginx role.

jpaulovic:
Move SSL configuration entirely to the nginx role.
Remove unnecessary nginx config file.


robled (dismissed):

cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -17,16 +17,6 @@ update_centers:
...
 # SSL certificate, otherwise a self signed certificate will be used.
 https_enabled: false
-# Local path to the SSL certificate to use in configuring HTTPS. This can be
-# overridden if you want secure access to Jenkins using your own SSL certificate.
-# By default, a self signed certificate created by the nginx role will be used.
-#jenkins_ssl_cert:
-# Local path to the SSL encryption key to use in configuring HTTPS. See the
-# comment on jenkins_ssl_cert for more information.
-#jenkins_ssl_key:
-# The version of Jenkins to install from the repository. The only use for
```
greg-hellings: Why have you removed this comment?

Also: why have you removed the jenkins_ssl_* variables altogether?

greg-hellings (changes requested):

cinch/roles/jenkins_master/defaults/main.yml
jpaulovic: Because since nginx became an integral part of cinch and it is taking care of the HTTP(S) termination, I left the termination configuration up to the nginx role to remove redundancy, as @robled implied in comments to the previous PR. I believe that makes the code cleaner while not affecting functionality.

greg-hellings: OK, but I think you excised too much here - you've also removed the comment that explains the jenkins version variable below.

jpaulovic: Ah, you're right. Fixed.

cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -17,16 +17,9 @@ update_centers:
...
-# overridden if you want secure access to Jenkins using your own SSL certificate.
-# By default, a self signed certificate created by the nginx role will be used.
-#jenkins_ssl_cert:
-# Local path to the SSL encryption key to use in configuring HTTPS. See the
-# comment on jenkins_ssl_cert for more information.
-#jenkins_ssl_key:
 # The version of Jenkins to install from the repository. The only use for
 # this value is constructing the name of the RPM below, so there is no need
-# to override this just outside of changing RPM name
+# to override this just outside of changing RPM name.
```
greg-hellings: OK, now there's a LOT of noise here with periods added all throughout the place. Why is that?

jpaulovic: Consistency. Why do some sentences finish with a period and some not? (I guess because of different contributors)

greg-hellings: OK, well the majority seem to not have it at the end (which is consistent with other style guides I've seen). I prefer that over having them everywhere, especially as many of them are not full sentences. Rather than introduce more git noise, can you revert the commit you just pushed and force push over the branch with just restoring the comment at this point?

jpaulovic: OK, done

greg-hellings (approved):

robled (approved):

PR #119, opened 2017-05-26
added check_mode support to Jenkins configuration groovy scripts

robled:
* Added an optional jenkins_check_mode tag to only check the Jenkins
configuration


greg-hellings (approved):
This might be overkill, as nothing should be altered until you call ``jenkins.save()``, but it's good as-is.

PR #122, opened 2017-06-01
Install additional pip packages on Jenkins master

jpaulovic:


robled (approved):

greg-hellings (changes requested):
I'd like to avoid installing anything further through pip at the global namespace, please?

robled:
Certainly I agree with @greg-hellings with regards to installing things from **pip** system-wide, but having this option is nice for those times when something just isn't packaged for a distro and using a virtualenv is somehow not desired.  Is there a specific use case we are accounting for?

robled:
Per a recent meeting we've decided to not include this feature in cinch, as we don't currently have a use case for it.  In the event that a use case develops (or returns), then we can resurrect this PR.

PR #123, opened 2017-06-01
Add option to install CA certs from local files

jpaulovic:


robled (dismissed):

cinch/roles/certificate_authority/tasks/main.yml
```diff
@@ -1,7 +1,20 @@
...
+- name: install certificate authority from URL
   get_url:
     url: "{{ item }}"
     dest: /etc/pki/ca-trust/source/anchors/
+    owner: root
+    group: root
+    mode: 0644
     validate_certs: "{{ certificate_authority_validate_certs }}"
   with_items: "{{ certificate_authority_urls }}"
   notify: install certificate authority
```
greg-hellings, line 11: If we're forcing download to be as "root", let's add a "become: true" to this line.

greg-hellings: And the task below it.

greg-hellings (changes requested):

cinch/roles/certificate_authority/tasks/main.yml
robled: Is it OK to have site.yml take care of this for us?

https://github.com/RedHatQE/cinch/blob/master/cinch/site.yml#L15

IIRC, most things in cinch require root and we just usually add a **become: false** for the few places that don't need it.

jpaulovic: That's a good point, I don't see we would use become in any other roles, for example when installing packages with yum, where you also need root. I'd say @robled is right and it's unnecessary... OTOH it's true that if the role should be reusable, it would state that it needs root permissions for specific tasks. However, in that case, it would make sense to also change it in the other roles in the future.

robled: We've decided to go ahead and add **become: true** for the places that need it in spite of what is contained in site.yml or other places where this is not explicitly set.  This will be the standard from now on.

cinch/roles/certificate_authority/handlers/main.yml
```diff
@@ -1,2 +1,2 @@
 - name: install certificate authority
-  command: update-ca-trust
+  command: update-ca-trust force-enable
```
greg-hellings, line 2: What does this "force-enable" option do? On my version of update-ca-trust, that command is not document. The only documented command is either an empty command or the "extract" command.

greg-hellings (approved):

cinch/roles/certificate_authority/handlers/main.yml
jpaulovic: Yeah, this is very much RHEL-specific option to force the update...

PR #126, opened 2017-06-07
Issue28.b

greg-hellings:


cinch/library/jenkins_user_api.py
```diff
@@ -0,0 +1,49 @@
+from ansible.module_utils.basic import AnsibleModule
```
robled, line 1: Add some comments to this file.

cinch/library/jenkins_user_api.py
```diff
@@ -0,0 +1,49 @@
...
+from tempfile import NamedTemporaryFile
+import subprocess
+import os
+
+
+MYFILE = """
```
robled: If there's not a way to avoid embedding this code, let's add a comment explaining why so that no one else tries to excise it in the future.

cinch/library/jenkins_user_api.py
```diff
@@ -0,0 +1,49 @@
...
+            'cli_jar': {'default':
+                        '/var/cache/jenkins/war/WEB-INF/jenkins-cli.jar'},
+            'jenkins_url': {'default': 'http://localhost:8080'}
+        }
+    )
+    args = type('Args', (object,), module.params)
+    groovy = NamedTemporaryFile(delete=False)
+    groovy.write(MYFILE.format(args.user))
+    groovy.close()
+    process = ['/usr/bin/java',
```
robled: Probably unlikely, but we might want to set this as a module argument in the event that the java binary is stored elsewhere.

cinch/roles/jenkins_common/handlers/main.yml
```diff
@@ -4,3 +4,22 @@
...
+    key_file: "{{ jenkins_user_home }}/.ssh/id_rsa.pub"
+    state: present
+  become: true
+  ignore_errors: true
+  notify: restart Jenkins
+  # Errors ignored because the file might not exist on slaves, and it might not
+  # yet be configured with roles, etc if this is executed on first run. However,
+  # it becomes important to have this running during subsequent executions
+
+- name: retsart Jenkins
```
robled: typo

cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -43,6 +43,54 @@ jenkins_args:
...
+# ability to control the Jenkins master instance without authentication
+jenkins_security_enabled: false
+# An admin user who is configured to permit automated connections to the system
+# for things like running the Groovy scripts, connecting Swarm clients, etc.
+# This user is automatically assigned to the security role "admin" that gets
+# created
+jenkins_admin:
+  nickname: jenkins-admin
+  email: root@localhost
+  password: adminpassword
```
robled: A password of 'changeme' might be a little more self-documenting.

cinch/roles/jenkins_master/tasks/post_configure.yml
```diff
@@ -20,9 +15,24 @@
...
+    - script: "{{ lookup('template', role_path + '/templates/set_executors.groovy') }}"
+    - script: "{{ lookup('template', role_path + '/templates/set_usebrowser.groovy') }}"
+    - script: "{{ lookup('template', role_path + '/templates/set_slaveport.groovy') }}"
+    - script: "{{ lookup('template', role_path + '/templates/user.groovy') }}"
+      when: "{{ jenkins_security_enabled }}"
+    - script: "{{ lookup('template', role_path + '/templates/roles_and_ldap_auth.groovy') }}"
+      when: "{{ jenkins_security_enabled }}"
+  become: true
+  become_user: "{{ jenkins_user }}"
+  when: (item.when is not defined or item.when) | bool
```
robled: This when statement looks somewhat unusual to me.  For my own purposes I wouldn't mind a comment here.

cinch/roles/jenkins_master/templates/roles_and_ldap_auth.groovy
```diff
@@ -0,0 +1,177 @@
+import hudson.*;
```
robled, line 1: There's some long lines in this file.  As of right now the standard is 80 characters (I know that this is not ideal in the view of many (myself included), but until the standard is officially changed we might as well adhere to it).

robled: I'm seeing some mixed indents in this file.  Can we standardize on four spaces?  Perhaps I just don't understand how the parens in groovy affect indentation levels.

robled (changes requested):

cinch/library/jenkins_user_api.py
greg-hellings: Commented.

cinch/library/jenkins_user_api.py
greg-hellings: There is no way. Alas.

cinch/library/jenkins_user_api.py
greg-hellings: Good point. I forgot to pull that over from the other system.

Having a different JRE command is actually not at all unlikely as it's relatively common for OpenJDK and Oracle JDK to co-exist along with having multiple versions of the JDK installed.

cinch/roles/jenkins_master/tasks/post_configure.yml
greg-hellings: Oh, you mean that boolean logic that took me at least 7 tries to get right? Yeah, let me toss a comment on that one.

cinch/roles/jenkins_master/templates/roles_and_ldap_auth.groovy
greg-hellings: 80 (79, actually, because Guido is a dork) is only our specification for Python code. Yaml files we permit out to 100, and we had spoken about the lack of need for Groovy/Java to have such strict types of limitations. There were two lines in this file that were above 100 lines, so I've cut them down to under that limit, as we had talked about keeping it in the 100-110 limit for side-by-side diffs.

As for indents: they are consistent throughout this file. Block indents are one tab character, alignment of arguments within hanging indents for line continuation, etc, are all done with space. This gives a clean, manageable, and user-configurable tab stop width while also allowing proper alignment and hanging indents.

One of the beauties of Groovy is that indents are syntactically meaningless. This allows greater flexibility in achieving legibility with proper alignments and such. Can you point specifically to the indents that are bugging you, because this file reads to me exactly how I would expect it to read in Groovy. (And no, I'm not OK with settling on 4 space indents for languages other than Python)

robled: Seems like Groovy may need its own style guide (yet to be defined), so let's just leave this as-is for now.

robled (approved):

PR #129, opened 2017-06-12
Added Jenkins restart and diagnostics utility

robled:
* Created a new directory named 'playbooks' to store one-off automation
playbooks that should not be part of a role or referenced in site.yml

* Moved install-rhel7.yml to the new 'playbooks' directory and created a
symlink for backwards compatibility


cinch/playbooks/jenkins_restart_utility.yml
```diff
@@ -0,0 +1,137 @@
+# This playbook aims to gather diagnostics data in the event of
+# trouble with Jenkins, and then restarts Jenkins to keep operations running.
+# By default, only the diagnostics data is collected, but the Jenkins service
+# restart can be enabled by setting the 'restart_jenkins' variable to true:
+
+# Example:
+# ansible-playbook -u root --private-key /path/to/key -i '10.0.183.169', jenkins_restart_utility.yml -e "restart_jenkins=true"
```
zregvart: Does this mandate root access, or can it be run by a non-privileged user, for instance from `jenkins` account?

cinch/playbooks/jenkins_restart_utility.yml
```diff
@@ -0,0 +1,139 @@
...
+                remote_src: true
+              when: diag_path_creation|success and support_bundle_file is defined
+          when: jenkins_running.stat.isreg is defined
+          ignore_errors: true
+      become: true
+      become_user: jenkins
+
+    # always restart Jenkins if restart_jenkins is true
+    - name: restart Jenkins
+      service:
```
abraverm, line 144: Be careful using service in Docker containers: https://github.com/ansible/ansible-modules-core/issues/593

cinch/playbooks/jenkins_restart_utility.yml
```diff
@@ -0,0 +1,139 @@
...
+                content: "{{ jvm_memory_jstat.stdout }}"
+                dest: "{{ diag_path }}/jvm-jstat.txt"
+              when: diag_path_creation|success
+
+            # By default, the 'root' Unix user account is assumed to have the
+            # necessary SSH key to authenticate to Jenkins as an admin user who has
+            # the authorization to generate the support bundle.
+            - name: generate support bundle
+              command: >
+                timeout 120 java -jar
```
abraverm: /bin/java

cinch/playbooks/jenkins_restart_utility.yml
```diff
@@ -0,0 +1,139 @@
...
+              copy:
+                content: "{{ thread_dump.stdout }}"
+                dest: "{{ diag_path }}/jstack.txt"
+              when: diag_path_creation|success
+
+            # TODO: add the following package so that this task can work properly:
+            # java-1.8.0-openjdk-debuginfo
+            - name: generate JVM memory report (jmap)
+              command:
+                timeout 60 jmap -heap {{ jenkins_pid }}
```
abraverm: I suggest full path of jmap command

cinch/playbooks/jenkins_restart_utility.yml
```diff
@@ -0,0 +1,139 @@
...
+              # By default, strings from slurp are base64 encoded with newlines,
+              # and that must be removed
+            - name: set Jenkins PID for later use
+              set_fact:
+                jenkins_pid:
+                  "{{ slurped_jenkins_pid['content'] | b64decode | regex_replace('\n', '') }}"
+
+            - name: generate thread dump
+              command: >
+                timeout 60 jstack -l {{ jenkins_pid }}
```
abraverm: I suggest full path of jstack command

cinch/playbooks/jenkins_restart_utility.yml
```diff
@@ -0,0 +1,139 @@
...
+            dest: "{{ diag_path }}"
+            remote_src: true
+          with_items: "{{ jenkins_logs.files }}"
+          when: diag_path_creation|success
+
+        - name: grab extended diagnostics data if jenkins is currently running
+          block:
+            - name: find Jenkins PID
+              slurp:
+                src: "{{ jenkins_pid_path }}"
```
abraverm: see my note at line 19

cinch/playbooks/jenkins_restart_utility.yml
```diff
@@ -0,0 +1,139 @@
...
+- hosts: all
+  vars:
+    jenkins_pid_path: '/run/jenkins.pid'
+  tasks:
+    # This block ignores all errors to ensure that Jenkins is restarted on
+    # demand even if we can't gather diagnostics data.
+    - name: gather diagnostics data
+      block:
+        - name: check if Jenkins is running
+          stat:
```
abraverm: I had cases where pid didn't work as indicator - the file is managed by Jenkins and if it crashed or killed you will have pid file pointing on non existing process. I prefer using the systemctl status to check if the process is alive, because its maintained by Jenkins developers. The hacky option but most reliable is  to check if there is a process alive that has jenkins jar in its name.

cinch/playbooks/install-rhel7.yml
```diff
@@ -0,0 +1,121 @@
...
+
+    # This pip install should be non-editable, but the pip module in Ansible
+    # 1.8.4. does not support that flag
+    - name: install linch-pin from master branch instead of latest release from pypi
+      command: >-
+        "{{ python }}" "{{ venv_dir }}/bin/pip install" -U
+        git+https://github.com/CentOS-PaaS-SIG/linch-pin.git@master
+      when: (linchpin_master|bool)
+
+    - name: install python-krbV with pip to use kerberos with Beaker
```
abraverm, line 93: I suggest always to be bound to a specific version

cinch/playbooks/install-rhel7.yml
```diff
@@ -0,0 +1,121 @@
...
+      get_url:
+        url: https://bootstrap.pypa.io/get-pip.py
+        dest: "{{ temp_dir }}"
+
+    - name: install pip manually by running get-pip.py script
+      command: "{{ python }} {{ temp_dir }}/get-pip.py"
+      args:
+        creates: "{{ venv_dir }}/lib/python2.7/site-packages/setuptools"
+
+    - name: install cinch using pip
```
abraverm, line 79: I suggest always to be bound to a specific version

cinch/playbooks/install-rhel7.yml
```diff
@@ -0,0 +1,121 @@
...
+    - name: >-
+        delete existing virtualenv directory (disabled by default, override
+        with DELETE_VENV Jenkins job parameter or delete_venv playbook variable)
+      file:
+        path: "{{ venv_dir }}"
+        state: absent
+      when: (delete_venv|bool)
+
+    - name: create virtualenv
+      command: virtualenv --no-setuptools "{{ venv_dir }}"
```
abraverm, line 60: full path of the command

cinch/playbooks/jenkins_restart_utility.yml
robled: The service restart functionality requires root as you would expect by default in any distribution.  The diagnostics collection can be done by the jenkins user, except for a single case where root may own the SSH key needed to authenticate to Jenkins to run the "support bundle" collection.  I'll contact you in private for further details.

cinch/playbooks/jenkins_restart_utility.yml
robled: This was a suggestion by @olivergondza, and if possible I'd like to get his thoughts on this.  It may be the case that the diagnostics collection that depends on the Jenkins PID would be unavailable anyways in the case of a Jenkins crash.

cinch/playbooks/install-rhel7.yml
robled: This is older code that was not written for this PR but rather just moved around to follow a new path convention.

cinch/playbooks/install-rhel7.yml
robled: This is older code that was not written for this PR but rather just moved around to follow a new path convention.

cinch/playbooks/jenkins_restart_utility.yml
robled: This was not intended to be run in Docker containers, but I'll go ahead and make sure this part just doesn't execute in that case to be sure.

cinch/playbooks/install-rhel7.yml
robled: This is older code that was not written for this PR but rather just moved around to follow a new path convention.

cinch/playbooks/jenkins_restart_utility.yml
```diff
@@ -0,0 +1,139 @@
...
+          ignore_errors: true
+      become: true
+      become_user: jenkins
+
+    # always restart Jenkins if restart_jenkins is true
+    - name: restart Jenkins
+      service:
+        name: jenkins
+        state: restarted
+      when: restart_jenkins | default(false) and ansible_connection != 'docker'
```
greg-hellings: It is my personal preference to wrap parenthesis around uses of filters when these only affect part of the expression. Nowhere have I seen the order of operations for filter evaluation vs Boolean algebra defined, and wrapping the parenthesis just gives us a clearer notion that the "and" is connected to the whole term on the left

``when: ( restart_jenkins | default(false) ) and ansible_connection != 'docker'``

cinch/playbooks/jenkins_restart_utility.yml
```diff
@@ -0,0 +1,139 @@
...
+# By default, only the diagnostics data is collected, but the Jenkins service
+# restart can be enabled by setting the 'restart_jenkins' variable to true:
+
+# Example:
+# ansible-playbook -u root --private-key /path/to/key
+# -i '10.0.183.169', jenkins_restart_utility.yml -e "restart_jenkins=true"
+
+- hosts: all
+  vars:
+    jenkins_pid_path: '/run/jenkins.pid'
```
greg-hellings, line 12: Do you think it would be better to specify a default value for jenkins_restart here?

cinch/playbooks/jenkins_restart_utility.yml
greg-hellings: If stating the PID file isn't a guarantee that the process is up, perhaps we could change the value of the register on this task to make it clear that this is only testing for the existence of the file. So naming it something like ``jenkins_pid_file`` instead of ``jenkins_running`` could make that more clear.

cinch/playbooks/jenkins_restart_utility.yml
greg-hellings: The same goes for the other many variables that are given default values inline. Does the value of "vars" here mess with variable search order the way that the vars directory in a role does? Maybe we could clean up the use of filters if we had default values here?

cinch/playbooks/jenkins_restart_utility.yml
```diff
@@ -0,0 +1,139 @@
...
+
+              # By default, strings from slurp are base64 encoded with newlines,
+              # and that must be removed
+            - name: set Jenkins PID for later use
+              set_fact:
+                jenkins_pid:
+                  "{{ slurped_jenkins_pid['content'] | b64decode | regex_replace('\n', '') }}"
+
+            - name: generate thread dump (jstack)
+              command: >
```
greg-hellings, line 78: Here you use the ``>`` operator, but on line 81 and 92 you don't. Why is that?

cinch/playbooks/jenkins_restart_utility.yml
```diff
@@ -0,0 +1,139 @@
...
+# Example:
+# ansible-playbook -u root --private-key /path/to/key
+# -i '10.0.183.169', jenkins_restart_utility.yml -e "restart_jenkins=true"
+
+- hosts: all
+  vars:
+    jenkins_pid_path: '/run/jenkins.pid'
+  tasks:
+    # This block ignores all errors to ensure that Jenkins is restarted on
+    # demand even if we can't gather diagnostics data.
```
greg-hellings: Are you sure this comment is correct? The ``ignore_errors`` look like it's on the inner block to me. This block appears to only have ``become`` and ``become_user`` parameters. Am I misreading something?

greg-hellings (changes requested):

cinch/playbooks/jenkins_restart_utility.yml
olivergondza: I do not recall the original code but it was somewhat creative on figuring out the PID. I agree the pidfile value needs to be checked for existence. OTOH, if there is a way to ask systemd to "give us the pid of jenkins service or indicate it is not running", I would prefer that.

abraverm: I don't know of any pretty solution:

command: systemctl status jenkins

greg-hellings:
Can you modify the tests/lint.sh file to properly pick up the new playbooks?

cinch/playbooks/jenkins_restart_utility.yml
robled: There's no issue based on my testing with defining all the defaults in the **vars:** section, at least as far as overriding the default vars on the command line using the **-e** option.  Certainly with less filters lying around the code is a bit easier on the eyes, so I'm going to move that stuff into the **vars:** section.

cinch/playbooks/jenkins_restart_utility.yml
robled: You're right, I adjusted it for yamllint and messed up the indent.

cinch/playbooks/jenkins_restart_utility.yml
robled: This was an oversight, the **>** should be used for all those command bits.

cinch/playbooks/jenkins_restart_utility.yml
robled: Since we're removing the filter here and defining **restart_jenkins** in the **vars:** section, this problem goes away.

cinch/playbooks/jenkins_restart_utility.yml
robled: Indeed there is not a pretty solution here, even with systemd.  At least for the case of the Jenkins systemd service, there's not an easily parseable way to get the last PID without some serious shell nonsense that I'd like to avoid.  See the output of these commands if you're curious:

```
systemctl show jenkins --no-pager
systemctl status -l jenkins --no-pager
```

In the end, we just need to *try* to grab the diagnostics data, and we need the PID to do that, and if we can't get that data, then at least we can restart the service for the user.  The only other alternative I can think of here is to use something like `pgrep -f jenkins`, but what happens if some other process has **jenkins** in the name there?  Not great either.  I think that using **/run/jenkins.pid** as the best source of truth to grab the diagnostics data will have to be sufficient, and per @greg-hellings suggestion I will rename the task to better indicate that we're not definitively making a decision that Jenkins is actually running or not.

robled:
All review comments have been addressed to some degree - please re-review.

cinch/playbooks/jenkins_restart_utility.yml
```diff
@@ -0,0 +1,148 @@
...
+- hosts: all
+  vars:
+    jenkins_pid_path: '/run/jenkins.pid'
+    diag_path_prefix: '/var/lib/jenkins/cinch_diagnostics'
+    jenkins_log_dir: '/var/log/jenkins'
+    log_history_days: '-7d'
+    jstack_path: '/usr/bin/jstack'
+    jmap_path: '/usr/bin/jmap'
+    jstat_path: '/usr/bin/jstat'
+    java_path: '/usr/bin/java'
```
olivergondza, line 19: All these will be on PATH with `-devel` jdk installed. sounds like an overkill to use variables.

olivergondza:
Aside that I find it unfortunate the 2 separate operations of restart and collecting diagnostics are squashed into one, this looks good to me.

greg-hellings (approved):

cinch/playbooks/jenkins_restart_utility.yml
robled: I don't disagree, but I did find that some other parts of cinch do use vars for the full path to some java programs, so at least we're consistent.

robled:
The original script this was ported from had everything in one file, so that was carried over.  I'm going to deliver this to a customer soon, so if they would like it broken out for their use case we can look into that.

PR #130, opened 2017-06-13
nginx: Ensure installation of libsemanage-python 

jpaulovic:
In order to manage SELinux rules, this package must be installed (and apparently ain't in older RHEL versions (<=7.2)

### DIFF
--- a/cinch/roles/nginx/tasks/selinux.yml
+++ b/cinch/roles/nginx/tasks/selinux.yml
@@ -1,3 +1,8 @@
+- name: ensure libsemanage-python is installed
+  package:
+    name: libsemanage-python
+    state: installed
+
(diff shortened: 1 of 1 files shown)


cinch/roles/nginx/tasks/selinux.yml
```diff
@@ -1,3 +1,8 @@
+- name: ensure libsemanage-python is installed
+  package:
+    name: libsemanage-python
+    state: installed
```
tehsmyers: For consistency with the other roles, I think this should be `state: present`

tehsmyers (dismissed):

robled (approved):

greg-hellings (approved):

PR #133, opened 2017-06-26
linchpin 1.0.x compatibility

robled:
* Removed 'pushd/pop' work-around from wrappers.py now that linchpin has
the -w option on the CLI

* Updated docs links in wrappers.py to make the actual content easier to
find

* We don't need to raise an exception for unsupported linchpin commands
in cinchpin.  It's better to just print a normal error message for the
user.

* cinchpin now supports the new CLI options of linchpin and removes the
old ones

* Added some default linchpin CLI args that should be useful, such as
printing Ansible output, pointing to the correct workspace directory,
and setting the creds path to that of the workspace directory

* Added group_vars support to 'cinchpin init'

* Updated local_paths dictionary to account for the new group_vars
directory and to create the proper skeleton info

* Removed redundant places where we print an error and sys.exit(1).  We
can use sys.exit('msg') and get the same result.

* Vendored jenkins_script Ansible module because linchpin v1.0.x doesn't
depend on a recent enough version of Ansible to include the module
within Ansible itself

* Added the ability to optionally install the latest tip of
cinch+linchpin via the install-rhel7 Ansible playbook

* Added beaker-client package to install-rhel7 Ansible playbook

* Removed a couple of old work-arounds that from install-rhel7 Ansible
playbook that are no longer needed as of linchpin 1.0.x

* Added check_ssh role to teardown.yml.  This was missing before and was
a bug that was exposed under some conditions.

* Updated users.rst documention to reflect all the changes with linchpin
1.0.x and cinchpin

* JJB template now checks out example topologies into a subdirectory of
the Jenkins $WORKSPACE, which should be best practice

* Added $PROVIDER parameter to example JJB template, which allows for
easy switching between OpenStack and Beaker examples

* Removed work-arounds from JJB template which no longer apply due to
linchpin 1.0.x

* Updated JJB template to use the new cinchpin commands

* Added linchpin 'outputs' files as artifacts for the JJB template,
which are generated during provisioning and must be passed to the
teardown jobs

* The 'runtest' portion of the JJB template has a more meaningful
example other than just 'exit 0'

* Added outputs files to the copyartifact portion of the JJB template so
that teardown jobs can run as expected.  There's also a symlink
work-around for when some linchpin providers name their outputs file
with a different file extension.

* Bumped version in setup.py to 0.7.0, with a requirement for linchpin
1.0.1

### DIFF
--- a/cinch/playbooks/install-rhel7.yml
+++ b/cinch/playbooks/install-rhel7.yml
@@ -14,6 +14,8 @@
     temp_dir: "{{ venv_dir }}/tmp"
     python: "{{ venv_dir }}/bin/python"
     delete_venv: false
+    latest_tip: false
+    beaker_kerberos: true
 
   tasks:
     - name: fail if we are not running this playbook on RHEL7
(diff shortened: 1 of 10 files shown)


cinch/playbooks/install-rhel7.yml
```diff
@@ -76,25 +78,30 @@
...
 
     # This pip install should be non-editable, but the pip module in Ansible
     # 1.8.4. does not support that flag
-    - name: install linch-pin from master branch instead of latest release from pypi
+    - name: install latest tip of cinch+linch-pin instead of latest release from pypi
       command: >-
-        "{{ python }}" "{{ venv_dir }}/bin/pip install" -U
-        git+https://github.com/CentOS-PaaS-SIG/linch-pin.git@master
-      when: (linchpin_master|bool)
+        "{{ python }}" "{{ venv_dir }}/bin/pip" install -U
```
greg-hellings: Inclusion of the {{ python }} argument here is superfluous. By directly calling the particular executable out of $VENV/bin/, it automatically is keyed through its she-bang line to use the virtual environment's Python file. No harm here, just unnecessary code.

cinch/playbooks/install-rhel7.yml
```diff
@@ -76,25 +78,30 @@
...
+        git+https://github.com/CentOS-PaaS-SIG/linch-pin.git@develop
+        git+https://github.com/RedHatQE/cinch.git@master
```
greg-hellings: Does this syntax work if git is not installed? It might be better to use the full path to the automatic tarball file, like ``https://github.com/CentOS-PaaS-SIG/linch-pin/archive/develop.tar.gz``

greg-hellings (dismissed):

cinch/playbooks/install-rhel7.yml
robled: Probably not.  I've amended my commit to just use the tarballs delivered over HTTPS.

cinch/playbooks/install-rhel7.yml
robled: You're right.  I've removed this and tested it successfully.  Commit amended.

greg-hellings (approved):

PR #136, opened 2017-07-12
Correct Update Center URL

tehsmyers:
Update center URLs don't use the complete x.y.z version, they just use
x.y.

For comparison:

- http://updates.jenkins-ci.org/1.651/
- http://updates.jenkins-ci.org/1.651.1/

I also tweaked the `jenkins_version` var comment to match

### DIFF
--- a/cinch/roles/jenkins_master/defaults/main.yml
+++ b/cinch/roles/jenkins_master/defaults/main.yml
@@ -12,15 +12,19 @@
 # variable with those values.
 update_centers:
   - id: default
-    url: "https://updates.jenkins-ci.org/{{ jenkins_version }}/update-center.json"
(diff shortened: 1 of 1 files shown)


cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -12,15 +12,19 @@
 # variable with those values.
 update_centers:
   - id: default
-    url: "https://updates.jenkins-ci.org/{{ jenkins_version }}/update-center.json"
+    url: "https://updates.jenkins-ci.org/{{ jenkins_uc_version }}/update-center.json"
```
greg-hellings: Err, won't this value be undefined? I don't see it being added in here.

greg-hellings (approved):

Issue #135, opened 2017-06-28
Provide configurable ntp server option


scottlinux:
Hello, 

Relevant to my interests would be a way to specify an ntp server to optionally use in lieu of distro defaults.

Some configurable (but not required) var action to specify an ntp server would be awesome.

![great](https://user-images.githubusercontent.com/2176591/27652247-08ff2c96-5c09-11e7-8101-3bf42daaef74.png)

greg-hellings:
I believe that @jpaulovic has this on his radar, as it is relevant to his interests as well. If you have the time and skill for a PR, we'd be happy to include that in the NTP role.

scottlinux:
Thanks - I'll submit a PR for review.

PR #137, opened 2017-07-17
roles/ntp: add support for configurable ntp servers

scottlinux:
- New `ntp_servers` override to specify optionally desired ntp server config.

- Example and brief note in group_vars/all

- If `ntp_servers` is not defined, distro defaults are used.

For https://github.com/RedHatQE/cinch/issues/135

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 v 0.7.1 (30 June 2017)
 - Updated Jenkins repo URL
 - Added ability to exclude changes to Jenkins plugins+configuration
+- Added configurable NTP server support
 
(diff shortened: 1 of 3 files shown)


greg-hellings (changes requested):
Can we get a comment in the ChangeLog for this? That file was added after 0.7.0 was released and little updates like this would be nice to include there for release-time publication.

scottlinux:
Ok thanks - I'll update the PR with a ChangeLog bit added.

scottlinux:
Comment added in Change log

greg-hellings (dismissed):

robled (approved):

PR #141, opened 2017-07-25
Ability to install packages on both master and slave

tehsmyers:
These tools are all handy for diagnosing issues and the general
administration of both jenkins masters and slaves.

I'm honestly not sure if this is the best place for them, but do think that these packages are useful to have on both jenkins masters and slaves.

The vim install fails on f25, but so does the openjdk debuginfo line in the preceeding task, so again I'm not really sure what, if anything, to do about it.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,10 @@ v 0.7.1 (30 June 2017)
 - Updated Jenkins repo URL
 - Added ability to exclude changes to Jenkins plugins+configuration
 - Added configurable NTP server support
+- Added ability to install packages on both master and slave with single var.
+  This renames the `additional_packages` var to `jmaster_extra_rpms` to be
+  consistent with the existing `jslave_extra_rpms`, and adds a new var called
+  `extra_rpms` for use in the `jenkins_common` role.
 
 v 0.7.0 (30 June 2017)
 - Baseline
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,4 @@
-v 0.7.1 (30 June 2017)
+v 0.7.1 (next release)
 - Updated Jenkins repo URL
 - Added ability to exclude changes to Jenkins plugins+configuration
 - Added configurable NTP server support
--- a/cinch/roles/jenkins_common/defaults/main.yml
+++ b/cinch/roles/jenkins_common/defaults/main.yml
@@ -12,3 +12,6 @@ jenkins_user_uid: 1000090000
 # (relative or full) to files containing separate public keys.
 # e.g.: ['mykey.pub', 'otherkey.pub']
 jenkins_authorized_keys: []
+# A list of additional packages that you wish to install on both masters and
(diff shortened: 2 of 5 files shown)


greg-hellings:
Eww, tmux. #screenforever

Installing vim on f25 barebones systems directly became broken a while back due to some circular dependencies around vim-minimal. If you upgrade vim-minimal to latest on the systems, then you can install the vim package. But that's not too important, as f25 is now back to maintenance mode with the release of f26. Plus, Fedora is only a "nice to have" and not a hard requirement for our installers.

jpaulovic:
Just my $0.02:
I have to admit I don't really like the idea of installing these packages by default.
It's not an integral part of cinch, nor Jenkins. Furthermore, different sysadmins find different packages useful for their work and they already have the ability to install these packages via the additional_packages variable.

tehsmyers:
`additional_packages` is only for the [jenkins_master](https://github.com/RedHatQE/cinch/blob/master/cinch/roles/jenkins_master/tasks/install.yml#L37-L41) role, and these packages would be useful on both masters and slaves. There is also the `jslave_extra_rpms` list used in [jenkins_slave](https://github.com/RedHatQE/cinch/blob/master/cinch/roles/jenkins_slave/tasks/main.yml#L6-L9), so I'm thinking what might make sense is to add `additional_packages` to jenkins_common, and rename `additional_packages` in `jenkins_master` to `jmaster_extra_rpms` for consistency with the `jenkins_slave` role.

greg-hellings:
I'm not opposed to that proposed change. Perhaps just call it extra_rpms to keep the name in line with jslave/jmaster variables.

Like Jakub, though, I'm not sure that this belongs as a default set of installs across all of Cinch. As a way to placate everyone, you could create a new file in group_vars that defines additional_packages/extra_rpms to include this array. Then, including these packages would simply be a matter of tagging a host into the appropriate group during a Cinch run. That way, we can all benefit from the knowledge of that list, but we don't need to force the extra packages onto a system that doesn't want the debugging tools installed.

tehsmyers:
Yeah, I left that bit out, thanks for pointing it out. I'm definitely in favor of leaving these lists empty in the defaults and letting cinch callers handle this themselves.

tehsmyers:
PR topic updated, and this is the new commit message:

```
This renames the "additional_packages" var to "jmaster_extra_rpms" to
match the existing "jslave_extra_rpms" var, and also adds support for an
"extra_rpms" var that will install packages on both masters and slaves.
```

Is a changelog entry sufficient to deal with the potential backward-incompatibility implications of renaming `additional_packages` to `jmaster_extra_rpms`?

greg-hellings:
Yeah, I think that a ChangeLog entry is fine for now. I don't feel a need to be more explicit about such things before we reach a 1.0 milestone. Until then, everything is fair game to change it up.

robled (approved):

PR #145, opened 2017-08-16
Add max file size ulimit restriction

jpaulovic:


cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -139,13 +139,15 @@ jenkins_java_options:
...
+# These values change the default ulimits for file and process counts in
 # order to avoid failure of the system when these (typically very low) default
-# values are hit
+# values are hit. The last value limits maximum file size to prevent corrupt jobs
+# from using dangerously large amounts of disks space.
 jenkins_soft_nofile_ulimit: 4096
 jenkins_hard_nofile_ulimit: 4096
 jenkins_soft_nproc_ulimit: 4096
 jenkins_hard_nproc_ulimit: 30654
+jenkins_fsize_ulimit: "{{ ansible_memtotal_mb * 1024 }}"
```
robled: A comment explaining the rationale behind the var provided by Ansible as it relates to the application would be helpful here.

robled (changes requested):

cinch/roles/jenkins_master/defaults/main.yml
jpaulovic: Is the comment sufficient now?

robled: I know that there is a private discussion on the details of this change.  I would suggest that we include as much of that info as is possible in this comment to further explain the justification for this setting so that anyone looking over this in the future will have the maximum amount of context available.

jpaulovic: Fixed.

robled (approved):

PR #146, opened 2017-08-18
Removed packages formerly used for legacy provisioning tools

robled:
* Removed RHOS repos from group_vars

* Removed legacy Python requirements.txt file

* Removed a large number of RPM packages from the jenkins_master role
that are no longer needed and added dependencies needed by cinch for
Jenkins slave provisioning

* Removed various QE test tools from the jenkins_slave role.  These
types of packages should be provided by the user via Ansible variables
instead of being included in the role by default.


greg-hellings (approved):
Approved with great fervor.

PR #148, opened 2017-08-23
Multiple admins

greg-hellings:
Enables setting multiple Jenkins sids as admin users, including with CLI access.


cinch/library/jenkins_cli_user.py
```diff
@@ -40,56 +43,49 @@ def main():
...
-                    keys = keyroot.getiterator("authorizedKeys")
-                    if keys:
-                        for key in keys:
-                            if pub_key not in str(key.text):
-                                changed = True
-                                if key.text is None:
-                                    key.text = pub_key
-                                else:
-                                    key.text = str(key.text) + pub_key
+    with open(params.key_file) as key:
```
robled: Let's add a comment here that explains the intent of this block, and especially detail under what conditions we except things to be in a CHANGED state.

robled (dismissed):

robled (approved):

jpaulovic:
Tested, works. Thanks!

PR #138, opened 2017-07-20
Groovy scripts

greg-hellings:

PR #150, opened 2017-08-30
Refine thinBackup plugin configuration defaults

robled:
* wait_for_idle is problematic and is now disabled by default, see this
for more details:

https://github.com/RedHatQE/cinch/pull/138#discussion_r133901263

* Other defaults have been changed based on successful experience in a
production environment, but obviously these settings are all highly
environment-specific and should be adjusted accordingly.

* Note that this automation exists in the source tree but is entirely
disabled given that thinBackup has many limitations and users may
implement other, more comprehensive backup solutions at another level
such as on the underlying storage device.

### DIFF
--- a/cinch/playbooks/jenkins_backup.yml
+++ b/cinch/playbooks/jenkins_backup.yml
@@ -29,7 +29,7 @@
       diff_schedule: H 0 * * *
       max_sets: 2
       exclude: ''
-      wait_for_idle: true
+      wait_for_idle: false
       quiet_mode_timeout: 480
       build_results: false
       user_contents: true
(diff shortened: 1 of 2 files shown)


cinch/playbooks/jenkins_backup.yml
```diff
@@ -29,7 +29,7 @@
       diff_schedule: H 0 * * *
       max_sets: 2
       exclude: ''
-      wait_for_idle: true
+      wait_for_idle: false
       quiet_mode_timeout: 480
```
olivergondza, line 33: This has no effect with `waitForIdle` turned off. You might want to delete it or comment it in case we will ever bring this back.

olivergondza: Probably https://github.com/RedHatQE/cinch/pull/150/files#diff-75c7e2bf9ddde36511712d435e5cf082R53 should be triggered conditionally iff `waitForIdle` is true too.

greg-hellings (approved):

greg-hellings:
We're getting away from thin backup. Closing this.

PR #154, opened 2017-09-05
Compatibility for upgrade to Jenkins 2.x

robled:
* Allow upgrades via unpinning the 'jenkins' package #142

* Added conditional addition of the -remoting flag for the jenkins_cli
that's required for Jenkins versions >= 2.46.2

### DIFF
--- a/cinch/roles/jenkins_master/defaults/main.yml
+++ b/cinch/roles/jenkins_master/defaults/main.yml
@@ -21,6 +21,11 @@ https_enabled: false
 # to construct the name of the RPM below as well as update center URLS, so this
 # must include the complete "x.y.z" version, e.g. 1.651.3.
 jenkins_version: "1.651.3"
+# Temporarily remove the jenkins package from the package mananger pinning
+# configuration to allow upgrades of the jenkins package.  Note that other
+# packages blacklisted by the upgrade_blacklist variable will remain pinned via
+# the package manager configuration.
+jenkins_upgrade: false
 # Update center version based on jenkins_version, update center does not use
 # "z" bit of the version, for example 1.651.3 becomes 1.651. This split
 # incantation strips the '.z' version off of jenkins_version.
--- a/cinch/roles/jenkins_master/tasks/install.yml
+++ b/cinch/roles/jenkins_master/tasks/install.yml
@@ -1,3 +1,11 @@
+- name: unpin jenkins package for version upgrades
+  lineinfile:
+    dest: "{{ version_pin_file }}"
+    state: present
+    regexp: "^exclude=.*"
+    line: "exclude={{ upgrade_blacklist | difference(['jenkins']) | join(' ') }}"
+  when: jenkins_upgrade
+
 - name: install necessary RPM files
   package:
     name: "{{ item }}"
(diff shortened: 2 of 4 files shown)


cinch/roles/jenkins_master/tasks/install.yml
```diff
@@ -1,3 +1,10 @@
+- name: unpin jenkins package for version upgrades
+  lineinfile:
+    dest: "{{ version_pin_file }}"
+    state: absent
+    regexp: "^exclude=jenkins$"
```
greg-hellings: This will only find lines that are exclusively excluding Jenkins. I'm not certain about the pin file format - is there only one package per exclude line, or can a line read ``exclude=jenkins kernel glibc``?

robled: Yeah, you can provide a space-separated list.

"exclude List of packages to exclude from updates or installs. This should be a space separated list. Shell globs using wildcards (eg. * and ?) are allowed."

https://linux.die.net/man/5/yum.conf

greg-hellings: Let's try something more like this, then:
```
lineinfile:
  dest: "{{ version_pin_file }}"
  state: present
  regexp: "^exclude=(.*)jenkins(.*)$"
  backrefs: true
  line: "exclude=\1 \2"
```

According to the Ansible documentation, when you have ``backrefs`` set to true and ``state: present``, then things will only be changed if ``regexp`` matches an existing line. If the user has no exclude line that contains Jenkins, then the file will not be modified. However, if they do, this /should/ preserve the value of their exclusions while still allowing us to upgrade Jenkins.

As an alternative, if we want to force them to use the value of ``upgrade_blacklist``, we can go with something like
```
lineinfile:
  dest: "{{ version_pin_file }}"
  state: present
  regexp: "^exclude=.*"
  line: "exclude={{ upgrade_blacklist | difference(['jenkins']) | join(' ') }}"
```

This would replace whatever the exclude line is with the values currently configured in upgrade_blacklist (regardless of any manual edits) but omit the "jenkins" entry.

greg-hellings: Please note that I haven't tested the above plays, I'm just shooting from the hip, here.

robled: Thanks for the suggestions.  I had some trouble with finding a regex I was happy with to avoid extraneous spaces, and the second suggestion works well based on my testing.  I'm fine with the dependency on the upgrade_blacklist, and I've updated the comment on the jenkins_upgrade variable to account for this.

cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -21,6 +21,8 @@ https_enabled: false
 # to construct the name of the RPM below as well as update center URLS, so this
 # must include the complete "x.y.z" version, e.g. 1.651.3.
 jenkins_version: "1.651.3"
+# Remove jenkins package from the upgrade_blacklist variable to allow upgrades
```
greg-hellings: We might want to say "Temporarily ignore presence of jenkins in the upgrade_blacklist..." or such, in order to avoid confusion that this might edit the local Ansible variables on-disk or pull some other wonkiness.

greg-hellings (approved):

PR #161, opened 2017-09-11
removed ansible RPM package dependency - cinch generally runs via a v…

robled:
…irtualenv where ansible is installed


robled:
I'm going to re-do this on top of cinch 0.8.0.

PR #163, opened 2017-09-12
Fix jenkins url

greg-hellings:


cinch/roles/jenkins_master/tasks/ensure_up.yml
```diff
@@ -12,9 +12,16 @@
...
+- name: discover valid hostname
+  set_fact:
+    _jenkins_host: "{{ ansible_host | default(inventory_hostname | default('localhost')) }}"
+  tags:
+    - jenkins_check_mode
+
 - name: construct Jenkins URL
   set_fact:
-    _jenkins_url: "http{{ https_enabled | ternary('s', '') }}://localhost:{{ _listening_port }}"
+    _jenkins_url: "http{{ https_enabled | ternary('s', '') }}://{{ _jenkins_host }}:\
```
robled: Can we omit the **_listening_port** being concatenated to this string in the case of HTTPS configuration?  It looks like the port used for HTTPS is not something currently configurable in cinch, so in the case of a non-container Jenkins master the port will always be 443, and adding the port number means that the Jenkins global environment variable **Jenkins URL** will contain the port number (_cinch/cinch/roles/jenkins_master/templates/jenkins_root_url.groovy_), and this will change existing configurations that did not include the port number.

Alternatively we could consider altering the behavior of the _jenkins_root_url.groovy_ script.

greg-hellings: I'll test an option for this.

robled (dismissed):

robled (approved):

PR #164, opened 2017-09-13
remove default value for jenkins_user_uid, which is not needed for mo…

robled:
…st installations


cinch/roles/jenkins_common/defaults/main.yml
```diff
@@ -5,8 +5,11 @@ jenkins_ssh_keyfile: id_rsa
 # supported beyond the EOL support for their packages in the distribution of
 # choice
 java_version: 8
-# Needed for certain OpenShift instances
```
greg-hellings, line 8: Can you add this default value into the inventory that is used in generating the Docker containers?

greg-hellings (changes requested):

greg-hellings (approved):

Issue #147, opened 2017-08-22
Move python-pip/{{ python_pip_package }}  to jenkins_common


greg-hellings:
These packages are currently duplicated across jenkins_master and jenkins_slave roles.  Move them to jenkins_common and use the version from jenkins_master that is currently configurable.

greg-hellings:
This was resolved in #177

PR #177, opened 2017-09-16
Issue147

greg-hellings:
Addresses #147

### DIFF
--- a/cinch/group_vars/fedora
+++ b/cinch/group_vars/fedora
@@ -3,7 +3,7 @@ repositories:
     mirrorlist: "{{ fedora_mirrors }}repo=rawhide"
 
 _download_repositories:
-  - http://pkg.jenkins-ci.org/redhat-stable/jenkins.repo
+  - https://fedorapeople.org/~semyers/jenkins-rpm/jenkins1651.repo
 
 jenkins_master_repositories: []
 jenkins_slave_repositories: []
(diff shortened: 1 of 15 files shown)


cinch/group_vars/fedora
```diff
@@ -3,17 +3,14 @@ repositories:
     mirrorlist: "{{ fedora_mirrors }}repo=rawhide"
 
 _download_repositories:
-  - http://pkg.jenkins-ci.org/redhat-stable/jenkins.repo
+  - https://fedorapeople.org/~semyers/jenkins-rpm/jenkins1651.repo
```
robled, line 6: At this point I think we should default to Jenkins 2.60.3 for new deployments.

robled (changes requested):

cinch/group_vars/fedora
greg-hellings: That repo contains both 1.651.3 and 2.60.3. I only made this change because it is a much faster mirror for people who still want the older package and it was taking well over an hour to install from the upstream repo on Fedora.

robled: Understood.  Certainly there are other places where we would need to change to default to 2.60.3.   We should do that eventually, but perhaps not in this PR.

robled (approved):

cinch/group_vars/fedora
greg-hellings: There's a variable in the jenkins_master role's defaults: jenkins_version, I believe. That is what would need to switch. I'd like us to get to 2.60.3 before we call ourselves 1.0, for sure. Shall we open a ticket for that?

Issue #176, opened 2017-09-15
Cinch 0.8.x requires ansible >=2.3.0.0 because of new syntax in ntp/tasks


jaypoulz:
### Overview
After upgrading to cinch 0.8.x, cinch no longer works past the NTP override block. This is a regression from 0.6.0 that appears to be caused by a patch to add an optional NTP Server override. In my configuration, I am not supplying any overridden NTP servers, so defaults are assumed.

##### Versions Affected
0.8.0-present

##### Failing Configurations
linchpin==1.0.2 (*Note* there is an [issue](https://github.com/CentOS-PaaS-SIG/linchpin/issues/356) provisioning from beaker with version 1.0.3)
ansible==2.2.1.0
cinch>=0.8.0

##### Last Successful State Supporting Linchpin 1.0.2
linchpin==1.0.2
ansible== 2.2.1.0
cinch==0.6.0

##### Current Psuedo-Successful State
linchpin==1.0.3 (*Note* there is an [issue](https://github.com/CentOS-PaaS-SIG/linchpin/issues/356) provisioning from beaker with this version)
ansible>= 2.3.0.0
cinch>=0.8.0

**Note** If you already have a provisioned beaker system with a properly configured inventory, cinch will work properly from this point forward.

### Error Message
```
ERROR! 'name' is not a valid attribute for a Block

The error appears to have been in '/usr/lib/python2.7/site-packages/cinch/roles/ntp/tasks/main.yml': line 7, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- name: override ntp servers when ntp_servers is defined
  ^ here
```
### Analysis / Root Cause
It appears as though because of the syntax used in the new updated NTP file, cinch requires ansible [>=2.3.0.0](https://github.com/ansible/ansible/commit/fcb033b9e442e7e49921817729169a14ba53d99d). In essence, this means that the cinch >=0.8.0 releases are incompatible with linchpin 1.0.2 (or older).

greg-hellings:
#137 isn't the cause of this. If anything, it's the Ansible version that's been upgraded. It must no longer support name attributes on blocks.

jaypoulz:
[...]

PR #178, opened 2017-09-16
Bump dependency for linchpin and ansible

greg-hellings:
There are syntax changes we use that require the newest versions of
Ansible. The latest release of linchpin allows us to, again, work in
concert with their versions of Ansible, since they are now unbounded
open on the high end

Addresses #176


setup.py
```diff
@@ -29,9 +29,9 @@
     packages=find_packages(exclude=('library', 'bin')),
     include_package_data=True,
     install_requires=[
-        'ansible>=2.2.1',
+        'ansible>=2.3.2',
         'plumbum>=1.6.0',
-        'linchpin>=1.0.1'
+        'linchpin>=1.0.3'
```
robled: I'm in favor of depending on linchpin 1.0.2 since there is an issue with Beaker provisioning in 1.0.3:

https://github.com/CentOS-PaaS-SIG/linchpin/pull/358

robled (dismissed):

setup.py
greg-hellings: 1.0.2 is still pinned to Ansible==2.2.whatever, and our playbooks have actually been dependent on 2.3 for a while now (per the issue this branch is mentioned from). Not sure there's a happy resolution to this, at the moment.

robled: Is it just that NTP task with the named block that depends on Ansible 2.3 or is there more?

greg-hellings: That's the only portion I'm presently aware of. But I'm not certain - are there any modules we use that you know of?

robled (approved):

PR #186, opened 2017-10-05
Stop messing with the executors

### LINKED ISSUE #182: Disable or remove configuration of number of executors
This is a user setting and based on some feedback we've received we've decided not to make this part of configuration management that's handled by cinch.

greg-hellings:
Users have requested we stop managing the number of executors with cinch
and instead leave that as a manual option that users can manipulate for
themselves

Fixes #182


cinch/group_vars/cent7
```diff
@@ -9,10 +9,6 @@ _repositories:
 _download_repositories:
   - https://fedorapeople.org/~semyers/jenkins-rpm/jenkins1651.repo
 
-rpm_key_imports:
```
robled, line 12: This change is related to a separate issue.  We should either update the commit log and link to that issue or break this out into a separate PR.

robled (changes requested):

robled (dismissed):

robled (approved):

Issue #190, opened 2017-11-07
Jenkins slave creation depends on Ansible, but Ansible is not installed via the jenkins_master role


robled:
In the **install-rhel7.yaml** JJB template, we use Ansible to install cinch on the Jenkins master for the purpose of Jenkins slave provisioning.  However, the jenkins_master role doesn't actually install Ansible, so we need to provide that somewhere.  The installation playbook was originally written for Ansible 1.8.4 due to some legacy dependency concerns, so the best solution here is to set up a temporary virtualenv via the JJB template that will install Ansible 1.8.4 so that we can then install cinch for later use.

PR #191, opened 2017-11-08
Create a temporary venv for Ansible

atodorov:
Fixes: #190 

@robled can you review

### DIFF
--- a/jjb/install-rhel7.yaml
+++ b/jjb/install-rhel7.yaml
@@ -31,9 +31,15 @@
     builders:
       - shell: |
             #!/bin/bash -ex
+            virtualenv ansible.venv && source ansible.venv/bin/activate
+            pip install --upgrade setuptools pip
+            pip install 'ansible==2.4.1.0'
+
             export PYTHONUNBUFFERED=1 # Enable real-time output for Ansible
             ansible-playbook -i localhost, -c local \
                 "${WORKSPACE}/cinch/cinch/playbooks/install-rhel7.yml" \
                 -e delete_venv="${DELETE_VENV}" \
                 -e latest_tip="${LATEST_TIP}" \
                 -e beaker_kerberos="${BEAKER_KERBEROS}"
+
+            deactivate


jjb/install-rhel7.yaml
```diff
@@ -31,9 +31,16 @@
     builders:
       - shell: |
             #!/bin/bash -ex
+            virtualenv ansible.venv && source ansible.venv/bin/activate
+            pip install --upgrade setuptools
+            pip install 'ansible>=2.1'
```
robled: If this was tested successfully on Ansible 2.1, let's set this to lock the version with '==', since new versions of Ansible can break existing playbooks.

jjb/install-rhel7.yaml
```diff
@@ -31,9 +31,16 @@
...
+
             export PYTHONUNBUFFERED=1 # Enable real-time output for Ansible
             ansible-playbook -i localhost, -c local \
                 "${WORKSPACE}/cinch/cinch/playbooks/install-rhel7.yml" \
                 -e delete_venv="${DELETE_VENV}" \
                 -e latest_tip="${LATEST_TIP}" \
                 -e beaker_kerberos="${BEAKER_KERBEROS}"
+
+            deactivate
+            rm -rf ansible.venv/
```
robled: Let's remove this 'rm' here, since we have the workspace-cleanup plugin in use which will handle this for us.

robled (changes requested):
@atodorov - thanks for the PR!  Review comments have been added.

jjb/install-rhel7.yaml
atodorov: I think ansible 2.4.1.0 was installed on the master and I can pin the version to that but how do I know if everything is working as expected ? Ping me on IRC and I can provide you with internal link to this job.

atodorov: Note2: `jjb/code-coverage.yaml` also uses 'ansible>=2.1' syntax.

robled: If the playbook runs to completion without error, everything is OK as far as the installer goes.  I am just trying to avoid a new version of Ansible getting released and then breaking the installer, requiring us to have to make an unexpected change to this job.  I'm fine with any Ansible version, as long as it can run the playbook.  

I have reviewed your internal job for this (build 6), and the output looks good.

robled: I've done some testing today, and my current installer playbook is compatible with the latest release of Ansible, so please update this PR with the following Ansible version specification:

ansible==2.4.1.0

jjb/install-rhel7.yaml
```diff
@@ -31,9 +31,15 @@
     builders:
       - shell: |
             #!/bin/bash -ex
+            virtualenv ansible.venv && source ansible.venv/bin/activate
+            pip install --upgrade setuptools
```
robled: If we change this line to the following, the installation can use Python wheel packages, which speeds up the installation significantly:

pip install --upgrade setuptools pip

robled (approved):

PR #193, opened 2017-11-20
jenkins_restart_utility.yml: limit target host 

jpaulovic:
Modify jenkins restart utility to add ability to specify host and include arbitrary playbook when needed.

### DIFF
--- a/cinch/playbooks/jenkins_restart_utility.yml
+++ b/cinch/playbooks/jenkins_restart_utility.yml
@@ -5,9 +5,25 @@
 
 # Example:
 # ansible-playbook -u root --private-key /path/to/key
-# -i '10.0.183.169', jenkins_restart_utility.yml -e "restart_jenkins=true"
+# jenkins_restart_utility.yml
+# -e "restart_jenkins=true hostname=jenkins
+# domain=example.com"
 
-- hosts: all
+# Optionally include an arbitary role here
+# to run before the playbook is executed if needed"
+
+- hosts: localhost
+  connection: local
+  tasks:
+    - name: Add host to inventory
+      add_host:
+        name: "{{ hostname }}.{{ domain }}"
+        groups: "jenkins_master"
+
+- hosts: jenkins_master
+  roles:
+    - role: "{{ dependency_role | default() }}"
+
   vars:
     jenkins_pid_path: '/run/jenkins.pid'
     diag_path_prefix: '/var/lib/jenkins/cinch_diagnostics'


robled (dismissed):

robled:
The **ansible --syntax-check** part of the linting tests failed:

https://travis-ci.org/RedHatQE/cinch/jobs/304782440#L768

@jpaulovic - Can you confirm if the **ansible-playbook** command can run that playbook if the var is undefined?  Seems like it should be able to, but since **ansible-playbook --syntax-check** isn't happy with it, we should check.

jpaulovic:
@robled you are right. Should be fixed by the latest commit. Please note that the hostname and domain variables still need to be specified.

cinch/playbooks/jenkins_restart_utility.yml
```diff
@@ -5,9 +5,16 @@
...
 # Example:
 # ansible-playbook -u root --private-key /path/to/key
-# -i '10.0.183.169', jenkins_restart_utility.yml -e "restart_jenkins=true"
+# jenkins_restart_utility.yml
+# -e "restart_jenkins=true hostname=jenkins
+# domain=example.com"
+
+# Optionally include an arbitary role here
+# to run before the playbook is executed if needed"
+- hosts: "{{ hostname }}.{{ domain }}"
```
greg-hellings: I'm not so keen on forcing a host to be dynamically constructed out of variables, here. Why are you doing this instead of just referring to a group name, such as "run_pre_role" and letting inventory magic take care of things directly?

cinch/playbooks/jenkins_restart_utility.yml
```diff
@@ -5,9 +5,16 @@
...
+  roles:
+  - role: "{{ dependency_role | default() }}"
```
greg-hellings: Maybe this name could be improved? "dependency_role" might be better mnemonically as "pre_restart_role_name" or something like that.

greg-hellings (changes requested):

greg-hellings:
Can we get a comment in the change log for this?

cinch/playbooks/jenkins_restart_utility.yml
jpaulovic: One reason that comes to mind is that we need this to populate this dynamically from a ticketing system sending parameters via API to Ansible Tower. If we were to use inventories, we would need to include another wrapping playbook with a task which would use the add_host module to populate a proper inventory. I doubt this is wanted for upstream too and this is definitely an easier way to achieve this while not diverging from upstream.

greg-hellings: Can Tower not construct an inventory out of the argument?

greg-hellings:
Is this still important?

cinch/playbooks/jenkins_restart_utility.yml
jpaulovic: I have changed the behavior to use the add_host for constructing in-memory inventory, that should look nicer.

jpaulovic:
Can we force a re-run of the lint check?

greg-hellings:
It should re-run automatically when you pushed those changes. I'll double check.

greg-hellings (approved):

greg-hellings:
A manually triggered re-run completed successfully.

PR #197, opened 2017-12-13
Update default Jenkins version to 2.60.3

greg-hellings:
Update default version to the LTS we are currently using.


tehsmyers (approved):
Been doing this in my personal stuff for a few months now, and I continue to love that it is, in fact, this easy.

PR #198, opened 2017-12-16
Extra repositories

greg-hellings:
Adds support for the extra repositories, so users do not have to override the in-build repositories list in order to support their own custom repos as well.

### DIFF
--- a/cinch/roles/repositories/defaults/main.yml
+++ b/cinch/roles/repositories/defaults/main.yml
@@ -1,5 +1,11 @@
 # Override with an array of objects like you would find in playbook/group_vars/rhel7
-# if you want a different set of repositories than this package installs by default
+# if you want a different set of repositories than this package installs by default.
+# It is not recommended that you touch the "repositories" variable in your
+# variables, as that value is set to include repositories that are necessary for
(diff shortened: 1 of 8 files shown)


cinch/roles/repositories/defaults/main.yml
```diff
@@ -1,5 +1,11 @@
 # Override with an array of objects like you would find in playbook/group_vars/rhel7
-# if you want a different set of repositories than this package installs by default
+# if you want a different set of repositories than this package installs by default.
+# It is not recommended that you touch the "repositories" variable in your
+# variables, as that value is set to include repositories that are necessary for
+# master and slave software installation. Extra repositories is unused by this
```
robled: Change "Extra respositories" to the variable name - "extra_repositories" for clarity.  Should also say that "extra_repositories" is unused _by default_.

robled (changes requested):

cinch/roles/repositories/defaults/main.yml
greg-hellings: Updated

robled (approved):

Issue #189, opened 2017-10-30
Allow plugin lists to be fetched from a HTTP URL or git repo


robled:
Right now the plugin lists are embedded as files in the cinch repository.  For some use cases, you may want to track an external list of plugins for Jenkins master configuration management, and with the current system you must copy those external lists into your own repository to keep up to date.  We should consider adding an option to fetch those lists externally, so that the lists can be managed from a single source of truth rather than having to copy those external lists whenever a change is made.

robled:
This has been completed and released as part of cinch 1.0.0:

https://github.com/RedHatQE/cinch/releases/tag/v1.0.0

PR #200, opened 2018-01-04
Added example for bringing in plugin lists via a URL (#189)

robled:


greg-hellings (approved):
Weird about the comma vs newline behavior, but good to know!

Issue #187, opened 2017-10-24
Passing cinch an inventory that has a space in the absolute path causes SSH to fail


jaypoulz:
## Background
I am currently using cinch like so:
```
ansible-playbook --inventory inventories/${slaveTarget}.inventory \
  --extra-vars '{ "rpm_key_imports":[], "jenkins_master_repositories":[], "jenkins_master_download_repositories":[] }' \
   cinch/cinch/site.yml
```

When cinch unloads this inventory path, it get's something like this:
```
/inventories/ppc64le-slave.inventory
```

However, instead of using the relative path at the ssh step (for the purposes of the jenkins-slave role),
it instead unloads the full path:
```
/tmp/workspace/Multi-Arch QE/provision-multiarch-slave/inventories/ppc64le-slave.inventory
```
As you can see, this path has a space in it. 

## Resulting Command
The full command from the code rendered above results in the following command:
```
ansible -i /tmp/workspace/Multi-Arch QE/provision-multiarch-slave/inventories/ppc64le-slave.inventory ibm-p8-storage-03.khw.lab.eng.bos.redhat.com -m ping
```

## Error
This command fails with error:
```
ERROR! Extraneous options or arguments
```

## Comment
While my workflow might not match the traditional cinch use case, I suspect this bug affects the just calling `cinch` directly from the command line directly since the path expansion should be using the same mechanism.

robled:
This has been completed and released as part of cinch 1.0.0:

https://github.com/RedHatQE/cinch/releases/tag/v1.0.0

PR #201, opened 2018-01-08
Allow for spaces in SSH key file path in the check_ssh role (#187)

robled:
* Added single quotes to the SSH key file path so that the shell can find the
SSH key file if the path contains a space.


greg-hellings (approved):
Thanks for getting to this. I've been bitten by this several times and keep forgetting to address it.

PR #206, opened 2018-01-15
Fix groovy script so check_mode truly performs no modifications to the system

olivergondza:
Most of the groovy scripts was implemented in a way they performed modifications to Jenkins model but not persisted them to disk. This fixes groovy scripts implementing `check_mode` so it does not touch Jenkins model at all.

### DIFF
--- a/cinch/roles/jenkins_master/templates/init_backup.groovy
+++ b/cinch/roles/jenkins_master/templates/init_backup.groovy
@@ -7,6 +7,7 @@ def jenkins = Jenkins.getActiveInstance();
 def thinbackup = ThinBackupPluginImpl.getInstance();
 boolean changes = false;
 
+def check_mode = {{ ansible_check_mode|to_json }};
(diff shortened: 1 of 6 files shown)


cinch/roles/jenkins_master/templates/init_backup.groovy
```diff
@@ -27,58 +28,58 @@ def void changed(String field, def value) {
...
 }
 
 if( thinbackup.isBackupPluginArchives() != backupPluginArchives ) {
-	thinbackup.setBackupPluginArchives(backupPluginArchives);
+	if( !check_mode ) thinbackup.setBackupPluginArchives(backupPluginArchives);
 	changed("backup plugin archives", backupPluginArchives);
 }
 
-if( changes && !checkMode )
+if( changes && !check_mode )
```
olivergondza, line 84: The original variable was not declared anywhere in the repo.

greg-hellings (approved):

PR #208, opened 2018-01-15
Add support for jenkins_extra_plugins

tehsmyers:
I was attempting to use the functionality documented in
http://redhatqe-cinch.readthedocs.io/en/latest/config.html#jenkins-plugins,
specifically the "jenkins_extra_plugins" var, and it didn't appear to be
supported.


tehsmyers:
I tested this locally both with and without `--extra-vars='{"jenkins_extra_plugins": [{"plugins": ["ansicolor"],  url: "{{ omit }}"}]}'`, the travis master job seems to have failed for unrelated reasons.

tehsmyers:
Requested a review from @greg-hellings, since the odd mismatch between docs and functionality being addressed here means there may be historical reasons not to do this that I'm not aware of.

greg-hellings (approved):

PR #214, opened 2018-02-09
Added ability to copy the update_center_certificate from a remote

robled:
source.

### DIFF
--- a/cinch/roles/jenkins_master/tasks/plugins.yml
+++ b/cinch/roles/jenkins_master/tasks/plugins.yml
@@ -21,6 +21,7 @@
   copy:
     src: "{{ update_center_certificate }}"
     dest: "{{ jenkins_home }}/update-center-rootCAs/update-center.crt"
+    remote_src: "{{ update_center_certificate_remote_src }}"
     owner: jenkins
     group: jenkins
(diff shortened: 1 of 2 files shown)


cinch/roles/jenkins_master/tasks/plugins.yml
```diff
@@ -21,6 +21,7 @@
   copy:
     src: "{{ update_center_certificate }}"
     dest: "{{ jenkins_home }}/update-center-rootCAs/update-center.crt"
+    remote_src: "{{ update_center_certificate_remote_src | default(false) }}"
```
greg-hellings: Why have a default here instead of in the defaults.yml file?

robled: My thinking behind this was that the "parent" setting of update_center_certificate wasn't defined in defaults/main.yml either, and it felt strange to define the "child" setting there but not also the "parent", which is unused by default.  Perhaps we could just set update_center_certificate to false (instead of a file path) in defaults/main.yml and then we could add the update_center_certificate_remote_src setting in that file as well and it would make more sense?

greg-hellings (approved):

Issue #205, opened 2018-01-11
Add support for Kerberos anonymous auth


greg-hellings:
Add support for configuring Kerberus anonymous authentication to the Groovy scripts

PR #222, opened 2018-02-13
Prepare kerberos groovy script for anonymous access configuration

greg-hellings:
Includes script changes from @olivergondza and Ansible changes by myself.

Addresses #205


olivergondza:
@greg-hellings, it seems you have forgotten to push the ansible part.

greg-hellings:
@olivergondza Oops! Pushed that, now.

olivergondza (approved):
The code looks good to me

cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -168,6 +168,10 @@ jenkins_ldap:
   email_addr_attr: mail
   manager_dn: ''
   manager_password: ''
+# By default, the Kerberos plugin is configured to disallow anonymous login access.
```
robled: Doesn't Jenkins set the anonymous user with read-only authorization by default?  I think this is the primary use case here, in that you might not want to have log in just to perform a read-only operation like downloading a build log.  If this is true, perhaps we can update the comment to reflect what this is useful for and how it works by default.

robled (changes requested):

cinch/roles/jenkins_master/defaults/main.yml
greg-hellings: By default there are no permissions associated with anonymous access. Our default permissions set does create a read-only permission set that is set for the anonymous SID. But your are correct that "login" is a misnomer here. The problem is that the Kerberos plugin is blocking all access to anonymous user without this setting, before permissions even get checked. I'll update this comment to be more correct.

robled (approved):

PR #223, opened 2018-02-14
Allow nginx keytab and SSL certs to be copied from a remote source

robled:
NOTE:  please do not merge this until testing has been completed by @jpaulovic

### DIFF
--- a/cinch/roles/nginx/defaults/main.yml
+++ b/cinch/roles/nginx/defaults/main.yml
@@ -15,6 +15,12 @@ nginx_send_timeout: 120s
 ## variables unset by default
 httpd_no_error_pages: false
 https_enabled: false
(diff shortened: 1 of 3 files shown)


jpaulovic:
It seems this doesn't work for some reason. The vars are applied correctly to the target host (e.g. jmaster_extra_rpms are installed), but the correct certificate is not picked up, the copy task is skipped entirely and a self-signed ssl cert is generated.

cc @greg-hellings

jpaulovic:
Note: This seems to be a behavior of Ansible (bug?) - using with_first_found doesn't look for the files on remote filesystem even if remote_src is True.

Issue filed: https://github.com/ansible/ansible/issues/36299

Workaround suggested in #224

PR #224, opened 2018-02-16
Allow keytab and SSL certs to be sourced from a remote host

jpaulovic:
Due to a bug/feature (?) described in ansible/ansible#36299 it is not possible to apply #223. This approach works around the issue.


greg-hellings (approved):
LGTM

PR #226, opened 2018-02-23
Stop using my fedorapeople mirror

tehsmyers:
Now that we're not focused on 1.651, we can return to using the normal
jenkins mirrors for downloading jenkins. The slow mirror problem only
affected old releases that jenkins was no longer distributing out to all
of its mirrors, this does not affect more recent LTS releases like 2.60
and 2.89.


greg-hellings (approved):
Once CI completes, we can merge this.

tehsmyers:
> Once CI completes, we can merge this.

done!

PR #229, opened 2018-04-09
adds the ability to select which user to execute swarm service as

ryankwilliams:
Currently each time cinch configures and connects a Jenkins slave to its Jenkins master via swarm. The user used to start the swarm service is jenkins.

Within the jenkins_slave role, there is a default variable `run_swarm_as_jenkins`. Which currently is not implemented. Some environments using cinch to configure Jenkins slaves may wish to connect the slave to the master using a different user other than jenkins. i.e. root user. This PR provides the ability for the user to override this setting.

I changed the default value for `run_swarm_as_jenkins` from `false` to `true`. This was needed to not break existing cinch users. If this was not changed (kept as `false`), then current cinch configuration files would start the swarm service as root user.

i.e.

### Execute swarm as jenkins (default setting)
Does not need to be defined by the user.
```yaml
run_swarm_as_jenkins: true
```

### Execute swarm as root user (user defined setting)
```yaml
run_swarm_as_jenkins: false
```

### DIFF
--- a/cinch/roles/jenkins_slave/defaults/main.yml
+++ b/cinch/roles/jenkins_slave/defaults/main.yml
@@ -12,10 +12,6 @@ jswarm_execs: 10
 jswarm_extra_args: ''
 # Directory where the swarm will live and execute
 jswarm_home: "{{ jenkins_user_home }}"
-# User which will execute swarm service
-jswarm_user: "{{ jenkins_user }}"
-# Whether the swarm should execute as the jenkins user. False means it will exec as root
-run_swarm_as_jenkins: true
 # Path to Java interpreter to use to run the Swarm client
 jenkins_java_cmd: /usr/bin/java
 # Command line args to the Java executable, such as heap size, etc
--- a/cinch/roles/jenkins_slave/defaults/main.yml
+++ b/cinch/roles/jenkins_slave/defaults/main.yml
@@ -12,8 +12,10 @@ jswarm_execs: 10
 jswarm_extra_args: ''
 # Directory where the swarm will live and execute
 jswarm_home: "{{ jenkins_user_home }}"
+# User which will execute swarm service
+jswarm_user: "{{ jenkins_user }}"
 # Whether the swarm should execute as the jenkins user. False means it will exec as root
-run_swarm_as_jenkins: false
+run_swarm_as_jenkins: true
 # Path to Java interpreter to use to run the Swarm client
 jenkins_java_cmd: /usr/bin/java
 # Command line args to the Java executable, such as heap size, etc
--- a/cinch/roles/jenkins_slave/tasks/main.yml
+++ b/cinch/roles/jenkins_slave/tasks/main.yml
@@ -57,14 +57,6 @@
   set_fact:
     swarm_retry_timer: 10
 
-# The default user to execute swarm service is jenkins. Users can override this
-# by setting the 'run_swarm_as_jenkins' variable to false. This will tell
-# cinch to execute the swarm service as root user.
-- name: set swarm to run as root user
-  set_fact:
-    jswarm_user: root
-  when: run_swarm_as_jenkins == false
-
 - name: upload swarm config file
   template:
     src: sysconfig_jenkins_swarm
--- a/cinch/roles/jenkins_slave/tasks/main.yml
+++ b/cinch/roles/jenkins_slave/tasks/main.yml
@@ -57,6 +57,14 @@
   set_fact:
     swarm_retry_timer: 10
 
+# The default user to execute swarm service is jenkins. Users can override this
+# by setting the 'run_swarm_as_jenkins' variable to false. This will tell
+# cinch to execute the swarm service as root user.
+- name: set swarm to run as root user
+  set_fact:
+    jswarm_user: root
+  when: run_swarm_as_jenkins == false
+
 - name: upload swarm config file
   template:
     src: sysconfig_jenkins_swarm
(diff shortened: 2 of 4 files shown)


cinch/roles/jenkins_slave/tasks/main.yml
```diff
@@ -57,6 +57,14 @@
...
     swarm_retry_timer: 10
 
+# The default user to execute swarm service is jenkins. Users can override this
+# by setting the 'run_swarm_as_jenkins' variable to false. This will tell
+# cinch to execute the swarm service as root user.
+- name: set swarm to run as root user
+  set_fact:
+    jswarm_user: root
+  when: run_swarm_as_jenkins == false
+
```
greg-hellings: We should avoid using `set_fact` to interact with user input variables as it is so high in the precedence chain as to render user input almost meaningless. Let's ditch `run_swarm_as_jenkins` entirely.

Also, what's wrong with the `jenkins_user` variable name? It is used extensively throughout this role. Either we should change it throughout this file, or we should stick with the `jenkins_user` variable which can already be overridden when executing against Jenkins slave hosts.

cinch/roles/jenkins_slave/defaults/main.yml
```diff
@@ -12,8 +12,10 @@ jswarm_execs: 10
 jswarm_extra_args: ''
 # Directory where the swarm will live and execute
 jswarm_home: "{{ jenkins_user_home }}"
+# User which will execute swarm service
+jswarm_user: "{{ jenkins_user }}"
 # Whether the swarm should execute as the jenkins user. False means it will exec as root
-run_swarm_as_jenkins: false
+run_swarm_as_jenkins: true
```
greg-hellings: Let's delete this value entirely.

cinch/roles/jenkins_slave/defaults/main.yml
```diff
@@ -12,8 +12,10 @@ jswarm_execs: 10
 jswarm_extra_args: ''
 # Directory where the swarm will live and execute
 jswarm_home: "{{ jenkins_user_home }}"
+# User which will execute swarm service
+jswarm_user: "{{ jenkins_user }}"
```
greg-hellings: Let's change this to `jswarm_user: "{{ jenkins_user | default('root') }}"`, if we maintain it at all.

greg-hellings (changes requested):
Overall, I'm not sure what you're accomplishing here by changing the name of the variable from `jenkins_user` to `jswarm_user`.  It's already possible to accomplish everything you're trying to do by defining `jenkins_user` to the name of the user you want to run the Swarm process as.

Ideally this PR would just delete the unused and uninteresting `run_swarm_as_jenkins` argument that is completely unreferenced anywhere in the role outside of `defaults/main.yml`. Then, you can continue to define `jenkins_user` in your group variables for the `jenkins_slave` hosts to be the user you want.

Am I missing something here?

ryankwilliams:
@greg-hellings I first tried setting the `jenkins_user` in my group_vars file `all` but never saw cinch taking my value due to it was getting overridden is my assumption? That is when I found the `run_swarm_as_jenkins` and thought it would be useful to have as a variable users can define in cases they do not want swarm to start as jenkins user.

I just tried a test of creating the correct group_vars file `jenkins_slave` and defined the `jenkins_user: root`. Cinch started to run and eventually failed when it was running the `jenkins_common` role. It was failing when it was attempting to create the jenkins user. It fails due to the current user is already the user that was defined, in this case root.

```bash
TASK [jenkins_common : create jenkins user] ************************************
fatal: [172.0.0.1]: FAILED! => {"changed": false, "msg": "usermod: user root is currently used by process 1\n", "name": "root", "rc": 8}
	to retry, use: --limit @/home/rywillia/Projects/rywillia/cinch/cinch/site.retry
```
If this is the approach you would like to take, the following change to the `jenkins_common` role would be needed in order for cinch to run successfully through this role when the `jenkins_user` as root.

```bash
diff --git a/cinch/roles/jenkins_common/tasks/main.yml b/cinch/roles/jenkins_common/tasks/main.yml
index 3cc372d..fd9bb13 100644
--- a/cinch/roles/jenkins_common/tasks/main.yml
+++ b/cinch/roles/jenkins_common/tasks/main.yml
@@ -33,6 +33,7 @@
     home: "{{ jenkins_user_home }}"
     shell: /bin/bash
     uid: "{{ jenkins_user_uid | default(omit) }}"
+ when: jenkins_user != ansible_user
```
How do you feel about this?

cinch/roles/jenkins_slave/defaults/main.yml
ryankwilliams: I will update the PR removing this value.

cinch/roles/jenkins_slave/tasks/main.yml
ryankwilliams: Sounds good, that makes sense. I will remove this from the PR.

greg-hellings:
@rywillia Maybe we could do `when: jenkins_user != ansible_user and jenkins_user != 'root'`? Because most of the time a user *shouldn't* be connecting as root. But it does happen depending on the cloud environment they're leveraging.

ryankwilliams:
@greg-hellings Yes I think that would be a good addition to the `when` statement. The case of users wanting to connect as root should be slim based on what you mentioned. I will update the PR based on our discussion.

greg-hellings (approved):

greg-hellings:
The erroring tests are related to something that's gone weird in the Apt repositories. I've tested locally and this seems fine.

PR #231, opened 2018-06-17
Fix ShellCheck

abraverm:
Shellcheck is not used at any point and the installation of it in Travis fail. Removing it fixed Travis CI: https://travis-ci.org/abraverm/cinch/builds/393286942

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,11 @@ python:  # Only one version for now
   - "2.7"
 cache: pip
 
+# Installation of the system
+addons:
+  apt:
+    packages:
+      - shellcheck
 before_install:
   - travis_retry pip install -U pip
   - pip --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,13 +7,6 @@ python:  # Only one version for now
   - "2.7"
 cache: pip
 
-# Installation of the system
-addons:
-  apt:
-    sources:
-      - debian-sid
-    packages:
-      - shellcheck
 before_install:
   - travis_retry pip install -U pip
   - pip --version


greg-hellings:
It is intended to be used in the tox lint target. There are still a good number of shell scripts which ought to be subject to its checking. It would be better to figure out what's going on in Travis and fix that then enable the checks.

abraverm:
I agree that shellcheck is good to have but its not currently used and it breaks the CI which prevents other work getting done. Fixing shellcheck for the sake of having it without using is pointless, and utilizing shellcheck is nice to have but not in priority. We can track the effort of shellcheck in another ticket.

scottlinux:
I am glad to look at travis to see what's going on

abraverm:
@scottlinux here what I found: https://github.com/koalaman/shellcheck/wiki/TravisCI
Also considering https://github.com/koalaman/shellcheck#travis-ci :

> Travis CI has now integrated ShellCheck by default, so you don't need to manually install it.

I don't know if we need to actually try to install it...

scottlinux:
Ok it looks like the issue was attempting to pull in a sid shellcheck in an ubuntu trusty travis, and perhaps sid has drifted too far in dependencies vs. trusty. (Error down below)

shellcheck is in trusty-backports so likely can get rid of the debian-sid part and just apt install it if needed?
```
addons:
  apt:
    packages:
      - shellcheck
```


```Installing APT Packages (BETA)
$ export DEBIAN_FRONTEND=noninteractive
8.17s$ sudo -E apt-get -yq update &>> ~/apt-get-update.log
1.01s$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install shellcheck
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:
The following packages have unmet dependencies:
 libc6-dev : Breaks: binutils (< 2.26) but 2.24-5ubuntu14.2 is to be installed
E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages.
apt-get.diagnostics
apt-get install failed```

abraverm:
@scottlinux thanks for looking into it, seems to work: https://travis-ci.org/abraverm/cinch/jobs/393643111

simzacks (approved):

PR #232, opened 2018-06-21
Inspec

abraverm:
Once https://github.com/RedHatQE/jenkins-profile/pull/1 will be merged, I will update the PR to use the main repository of jenkins-profile.

This enables Inspec testing for Cinch that will verify the end deployment mets the requirements. This is initial setup and doesn't include all required testing but enough to get started.


scripts/master.sh
```diff
@@ -26,7 +26,8 @@ ansible -i /dev/null \
 		detach=true \
 		command='/usr/lib/systemd/systemd \
 		--system' \
-		privileged=true"
+        capabilities=SYS_ADMIN \
```
greg-hellings: Check the white space indenting here. These lines don't match the ones above them.

tests/profile.yml
```diff
@@ -0,0 +1,161 @@
...
+    value:
+      - 'true'
+  - key: "hudson/securityRealm/@class"
+    value:
+      - 'hudson.security.HudsonPrivateSecurityRealm'
+  - key: hudson/slaveAgentPort
+    value:
+      - '50000'
+
+jenkins_plugins:
```
greg-hellings: This list will change for every different version of Jenkins. Are we sure we want to hard-code it into the test instead of base it off the version of Jenkins in an external file, the way we do with the main plugin list? This could get to be a very noisy file, otherwise.

greg-hellings (changes requested):

abraverm:
@greg-hellings, I changed what you asked, anything else?

greg-hellings (approved):
LGTM

Issue #227, opened 2018-03-06
Consider updating check_ssh role to use the new wait_for_connection module


robled:
**check_ssh** was written prior to the release of Ansible 2.3, which added a new module similar to what **check_ssh** does:

https://docs.ansible.com/ansible/latest/wait_for_connection_module.html

greg-hellings:
Fixed by #235

PR #235, opened 2018-08-30
Modify check_ssh role to use wait_for_connection module

xding2020:
- **[Modified]** Replace check_ssh role with wait_for_connection module (#227 )
- **[Removed]** Add groovy scripts that configure Jenkins in following aspects. This action gives an option for user to automatically fix security warning for newly deployed/updated Jenkins, and controled by the variable `jenkins_advised_security`, which is default as false.
    - Disable CLI -remoting mode
    - Enable CSRF Protection
    - Enable Agent to master security subsystem
    - Disable deprecated protocols

Edit: Sorry for this late modification. For the advised Jenkins security configurations, though it also works with **jenkins_security_enabled:false**, I think we could leave them for Jenkins users to decide. Another reason is that, the security warnings are mainly caused by Jenkins version 2.89, while we are now moving to 2.121.

### DIFF
--- a/cinch/roles/check_ssh/defaults/main.yml
+++ b/cinch/roles/check_ssh/defaults/main.yml
@@ -1,3 +1,2 @@
-ssh_retries: 100
+ssh_timeout: 600
 ssh_retry_delay: 6
-host: "{{ hostvars[inventory_hostname]['ansible_host'] | default(inventory_hostname) }}"
--- a/cinch/roles/check_ssh/tasks/main.yml
+++ b/cinch/roles/check_ssh/tasks/main.yml
@@ -1,23 +1,5 @@
-- block:
-  - name: verify that configured SSH private key is a file and has permissions of 0600
-    local_action: file path="{{ ansible_ssh_private_key_file }}" state=file mode=0600
-    tags:
-      - skip_ansible_lint
-
-  - name: construct command line for SSH check
-    set_fact:
-      check_ssh_cmd: >-
-        ansible --ssh-extra-args='-o StrictHostKeyChecking=no'
-        -i '{{ inventory_dir }}/{{ (inventory_file | string | default('')) | basename }}'
-        {{ host }} -m ping
-
-  - name: check for SSH connectivity and authentication
-    local_action: shell {{ check_ssh_cmd }}
-    register: result
-    changed_when: "result.rc != 0"
-    until: result.rc == 0
-    retries: "{{ ssh_retries }}"
-    delay: "{{ ssh_retry_delay }}"
-    tags:
-      - skip_ansible_lint
+- name: wait for ssh connection
+  wait_for_connection:
+    sleep: '{{ ssh_retry_delay }}'
+    timeout: '{{ ssh_timeout }}'
   when: ansible_connection == 'ssh'


cinch/roles/jenkins_master/templates/advised_security.groovy
```diff
@@ -0,0 +1,24 @@
+import jenkins.model.*
```
greg-hellings: Please add semicolons to the end of statements throughout this file.

cinch/site.yml
```diff
@@ -1,8 +1,13 @@
 - name: wait for SSH
   hosts: all
   gather_facts: false
-  roles:
-    - check_ssh
+  tasks:
```
greg-hellings: This should simply have replaced the body of the `check_ssh` role. `tasks` should be avoided in top level playbooks, especially those with multiple parameters that might be (in this case, are) used repeatedly.

cinch/roles/jenkins_master/tasks/post_configure.yml
```diff
@@ -29,6 +29,10 @@
     # jenkins_security_enabled is false
     - script: "{{ lookup('template', role_path + '/templates/basic_security.groovy') }}"
       when: "{{ not jenkins_security_enabled }}"
+    # This option will automatically configure Jenkins to avoid security warning
+    # jenkins_advised_security is default as false
+    - script: "{{ lookup('template', role_path + '/templates/advised_security.groovy') }}"
+      when: "{{ jenkins_advised_security }}"
```
greg-hellings: What happens when this is set, but `jenkins_security_enabled` is false? Is there any problem with that permutation of this, or are these security settings independent of the ones set by `jenkins_security_enabled`?

greg-hellings (changes requested):

cinch/roles/jenkins_master/templates/advised_security.groovy
xding2020: @greg-hellings Sure.

cinch/roles/jenkins_master/tasks/post_configure.yml
xding2020: @greg-hellings That's a good point. I need to test and think about this. Also, this script will disable the CLI -remoting, do you think this will conflict our work because we may need CLI -remoting on.

cinch/site.yml
xding2020: @greg-hellings OK, I would change this. But curious about why tasks should be avoided in top level playbooks.

cinch/roles/jenkins_master/tasks/post_configure.yml
tehsmyers: I'm *fairly* sure that -remoting is needed, but am a little too distant from it to say for sure. @lyrch might know, though.

cinch/site.yml
greg-hellings: Overall it is considered best practices to wrap all tasks inside of rolls. Especially if those rolls could be useful to others down the line. Cinch is not currently configured to use or consume rolls from Ansible Galaxy, but I'd love for that to change in the future so it can get on board with the [oasis](https://github.com/oasis-roles/) project. Waiting for SSH is definitely one of those highly sought after and reusable components.

cinch/roles/jenkins_master/tasks/post_configure.yml
greg-hellings: The -remoting flag work was initially added by @robled  (who GitHub allows me to tag when the name is correct) when we moved to the 2.x series. It was needed for the jenkins-cli.jar file that was being used in a few places. I would defer to his knowledge on this matter to know if it is still needed and in use.

robled: To my knowledge *-remoting* is still needed for Jenkins slave connectivity, likely by *slave.jar*:

https://jenkins.io/projects/remoting/

cinch/site.yml
xding2020: @greg-hellings  OK. Just sometimes I am confused by whether to wrap a **single task** into a role or just put it to the top level. From my view, it's a little bit redundant to make it a role. But since the check_ssh have some params, I think it's OK.

cinch/roles/jenkins_master/tasks/post_configure.yml
xding2020: Also, I want to clarify that, this script is a personal idea to give an option for user to fix the security warnings automatically. But if conflicts with the project, or not the features you guys want, I think no need to add this.

robled: With the default setting as `false`, I see no reason not to add this to the project.  When Jenkins slave consumers are ready to migrate to whatever option doesn't need *-remoting*, the variable can be set to `true`.

greg-hellings (approved):
Approved. I'm re-running the failed Traivs task. It appears to have timed out. Hopefully that will complete cleanly this time.

greg-hellings:
OK, Travis isn't doing its thing correctly, but this should be merged anyway.

PR #240, opened 2018-11-21
[CID-3645] Fix Jenkins session timeout configuration being ignored

abraverm:
See https://projects.engineering.redhat.com/browse/CID-3573?focusedCommentId=1007032&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1007032

### DIFF
--- a/cinch/roles/jenkins_master/defaults/main.yml
+++ b/cinch/roles/jenkins_master/defaults/main.yml
@@ -80,7 +80,7 @@ jenkins_java_extra_options: []
 # comfortably without having to re-login.
 session_timeout: 480
 # Command line arguments for Jenkins
-jenkins_args: "--sessionTimeout={{ session_timeout }}"
+jenkins_args:
 # A list of additional packages you wish to install using the default package
 # manager.
 jmaster_extra_rpms: []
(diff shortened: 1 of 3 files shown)


cinch/roles/jenkins_master/defaults/main.yml
```diff
@@ -80,7 +80,7 @@ jenkins_java_extra_options: []
 # comfortably without having to re-login.
 session_timeout: 480
 # Command line arguments for Jenkins
-jenkins_args:
+jenkins_args: "--sessionTimeout={{ session_timeout }}"
```
greg-hellings: Currently this would allow a user to accidentally override their `session_timeout` value unintentionally by setting a custom value for `jenkins_args`. It might be better to construct something like this in the `vars/main.yml` file (placing it in `vars/main.yml` instead of here in `defaults/main.yml` would be important, as values in `vars` are difficult to override in the precedence order and should be used by values that are not intended to be user configurable):

```yaml
_jenkins_args: --sessionTimeout={{ session_timeout }} {{ jenkins_args }}
```

Then you can change the places that currently reference `{{ jenkins_args }}` in templates to instead reference `{{ _jenkins_args }}`. That prevents the user from accidentally overriding the behavior of `session_timeout` without realizing it.

greg-hellings (changes requested):

cinch/roles/jenkins_master/defaults/main.yml
abraverm: good point, thanks.  I will do that.

abraverm:
@greg-hellings, can you approve changes?

greg-hellings (approved):

greg-hellings:
Sorry, missed the email notice of your push. LGTM

PR #243, opened 2019-01-23
Specify ansible lint version

abraverm:


greg-hellings (approved):
I've also opened #244 to fix the lint issues when we get around to upgrading to `ansible-lint >= 4`

PR #248, opened 2019-02-04
Release 1.4.0

abraverm:


greg-hellings:
Normally we've been adding Changelog entries before a release (we used to try and add them with each PR, but eventually gave up on that). Do you want to continue that here, or just cut the release?

greg-hellings (approved):

